PR #1, opened 2024-11-05
[18.0][MIG] mail_tracking: Migration to 18.0

trisdoan:
### Note
- missing commits from 14.0 are ported in https://github.com/OCA/social/pull/1475

#### Backend
- `access_rights_uid` was removed from `_search()` changed in https://github.com/odoo/odoo/pull/171374/commits/21d023e6bc00793ac3b016cfdcafd05c9a81d95d
- `_prepare_outgoing_list` got extra param `mail_server` in https://github.com/odoo/odoo/pull/96008/commits/97ce0844ca0c3d531a7a2e2e47a867d89835b24e
- `_message_format` was converted to use store and `message_format_extras` was removed in https://github.com/odoo/odoo/commit/52df49eabb6e44ea6e0ed30b61e5a555af2e606a. As consequences, `_extras_to_store` is introduced
- `tools.ustr` is considered to be inefficient and flagged as deprecated in https://github.com/odoo/odoo/pull/173811/commits/64a5724e88de6532cf2a04e9243d16b079319ffd. Hence, is replaced by `exception_to_unicode` for the same effect.

#### Frontend
- `mail.messaging` service was removed in https://github.com/odoo/odoo/commit/683258b28107f44fba363e3b4e69f181cac3f80e
- `initMessagingCallback` is removed in https://github.com/odoo/odoo/commit/4f8087f03ad904b99f8f3580557a3ea744a97b76. So new entry point is created for store `onStarted` instead.
- mail.thread service was removed in  https://github.com/odoo/odoo/commit/65e5a7943de094fe538f2b978cb58c028a5694a1

### This changes
- Store helper was introduced in https://github.com/odoo/odoo/pull/150625/commits/c032ccfbaef91da0ee244f0e44e0e4e9d8d8d3fb. As consequences, failed data is now passed through Store.
   - As a side effect, it is also easier to do the search filter_failed(thread) with Store service
  
- As Odoo uses less Jquery, the module runs fine with vanilla. This impacts `mail_tracking/static/src/components/message_tracking/message_tracking.esm.js`

- Took this chance to
   - replace % format with f-string
   - Improve permission constraint
       - Context: This [commit](https://github.com/OCA/social/pull/966/commits/b4ab1600fe0b464937ff3dc06ff7cbf3a29afe3a) porting from 12.0
       - Approach is summarized as below:
           1. When do the searching, 
             - First, use the query to get result
             - Browse those id with the query + Remove any forbidden records
             - Return the accessible records in form of query, to continue the process
           2. When user read the records
             - The low-level method `_check_access` will discard forbidden records + raise exception
       - Added security check test

    - Resolve FIXME, which was added in https://github.com/OCA/social/pull/624/commits/75604439061755da9df64f658a9951a85e196185, then was resolved in https://github.com/odoo/odoo/pull/49999/commits/7fe02c93e410f8e9d11c89f77c651ca4b93810ff

### Result
discuss_sidebar

failed_search_panel

failed_message_review


ivantodorovich (approved):

mail_tracking/models/mail_message.py
```diff
@@ -59,30 +61,44 @@ def _compute_is_failed_message(self):
...
+        self.flush_model(
+            ["author_id", "mail_tracking_ids", "mail_tracking_needs_action"]
+        )
+        self.env["mail.notification"].flush_model(["mail_message_id", "res_partner_id"])
+        # retrieve failed tracking email
+        tracking_operator = "in" if operand else "not in"
+        failed_states = list(self.get_failed_states())
+        tracking_ids = self.env["mail.tracking.email"]._search(
+            [("state", f"{tracking_operator}", failed_states)]
+        )
```
TDu: This seems very costly for a production database with 10 or 100 thousand of record in mail.tracking.email, it will return a lot of records.
I like better the previous implementation below,
`(
                    "mail_tracking_ids.state",
                    "in" if value else "not in",
                    list(self.get_failed_states()),
                ),`

ivantodorovich: Indeed, I agree

chienandalu (commented):
As for comment https://github.com/OCA/social/pull/1476#issuecomment-2440797879

> Look like it gets the same issue in 17.0

No, it doesn't. You can try in a runboat. If don't remember it wrong there was a bus call in python that notified the state to the client  so we didn't need to reload the whole page risking to loose context like the breadcrumb path...

mmequignon (commented):
I agree with @TDu comment about the double search, it should be done in 1.

trisdoan:
~~DRAFT: working on reactivity of reviewed failed msg~~

trisdoan:
Hello @TDu, @ivantodorovich , @mmequignon, I updated the `_search_is_failed_message`

trisdoan:
Hello @chienandalu, 

I just added [this patch](https://github.com/OCA/mail/pull/1/files#diff-6578fb43e499661f0367351a6ab86d84b1fbb8c3a3f053e9c95e3091344784fb) + [use bus](https://github.com/OCA/mail/pull/1/commits/974b492fe527c202c002789157a44e6978db70f3#diff-c0e1179ee039233e47799e5b0e832110769850a395bc95ffc5d1e42aa5c23e2bR284-R289) to fix the reactivity both in Discuss and Chatter, here is result:

[mail_tracking_v18.webm](https://github.com/user-attachments/assets/02df0bec-1bc6-455a-881c-79291daf7287)


Yes, you'r right, it works in v17 but only in Chatter, it does not work in Discuss (I ran in gevent mode), is that the same on your local ?

[mail_tracking_v17.webm](https://github.com/user-attachments/assets/68d1fd35-b9e8-42f1-8161-8357a553b7f3)

mmequignon (approved):
fine by me

PR #3, opened 2024-11-12
[18.0][MIG] mail_composer_cc_bcc: Migration to 18.0

trisdoan:
### Changes in 18.0
- uid was forced in https://github.com/odoo/odoo/commit/9bc32a3e3cd7ee792a43451fa4458126062aff93, which made some keys are missing compared to standard '_notify_get_recipients'

### This changes
[[IMP] mail_composer_cc_bcc: refactor _generate_template_recipients](https://github.com/OCA/mail/pull/3/commits/3b652200d9041389231b97e2e964076a8f84a310)
- Took this chance to refactor this long method
    - It was completely overridden in previous migration, which is not necessary
    - Also, it results in `Invalid field 'partner_bcc_ids' on model 'mail.mail` in case of mass mailing


mail_composer_cc_bcc/models/ir_mail_server.py
```diff
@@ -0,0 +1,28 @@
...
+_logger = logging.getLogger(__name__)
+
+
+class IrMailServer(models.Model):
+    _inherit = "ir.mail_server"
+
+    def _prepare_email_message(self, message, smtp_session):
+        """
+        Define smtp_to based on context instead of To+Cc+Bcc
+        """
```
Jerem-b, line 17: If i choose to add no default user inside Bcc (from company). But then i add them manually within the composer, i got this error: Invalid email address. It is because of the native _prepare_email_message which checks if email address is inside 'email_to', 'email_cc', 'email_bcc'. However, Bcc addresses is empty (inside the headers instead). So smtp_to_list is empty too and the assert smtp_to_list, self.NO_VALID_RECIPIENT returns the error.

I tested with Odoo 17.4 which is not 18 but closely, and i was able to bypass the error by adding my Bcc addresses list inside the message before calling super().

```
# Find the value for 'X-Odoo-Bcc'
x_odoo_bcc_value = next((value for key, value in message._headers if key == 'X-Odoo-Bcc'), None)
# Add Bcc field inside message to pass validation
if x_odoo_bcc_value:
    message['Bcc'] = x_odoo_bcc_value
```

trisdoan: Hi @Jerem-b, indeed, thank you so much :pray: 

I just updated the code to include your contribution

hbrunn:
~no time for reviewing this right now, but it would make sense to add a commit here to make you the maintainer here too I think~

ignore, you did so already in v17

hbrunn (approved):

AEstLo:
LGTM
Tested and seems to work:
image

AEstLo (approved):

trisdoan:
Hello @pedrobaeza, could you help to merge this please?

fyi, module is currently maintained by me in repo OCA/social

PR #6, opened 2024-11-13
[18.0][MIG] mail_inline_css: Migration to 18.0

nguyenminhchien:
- Drop the dependency to `email_template_qweb`:
  - The object `mail.template` from core module `mail` has already supported qweb content via the field `body_html` (https://github.com/odoo/odoo/blob/18.0/addons/mail/models/mail_template.py#L63).
  - So, i adjust this module to drop the dependency to `email_template_qweb`.


ivantodorovich (approved):
Thanks!

hbrunn (approved):

PR #17, opened 2024-11-27
[18.0][MIG] mail_activity_done: Migration to 18.0

achulii:


mail_activity_done/models/res_users.py
```diff
@@ -0,0 +1,59 @@
+# Copyright 2018-22 ForgeFlow <http://www.forgeflow.com>
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
+from odoo import api, fields, models, modules
+
+
+class ResUsers(models.Model):
+    _inherit = "res.users"
+
+    @api.model
+    def _get_activity_groups(self):
```
hbrunn, line 10: this has very little to do with what [upstream](https://github.com/OCA/OCB/blob/18.0/addons/mail/models/res_users.py#L324) does by now.

Did you check if we need this override at all? Given this is pure ORM, I'd expect things to just work because done activities are nonactive.

mail_activity_done/models/mail_activity.py
```diff
@@ -0,0 +1,92 @@
...
+        )._action_done(feedback=feedback, attachment_ids=attachment_ids)
+
+
+class MailActivityMixin(models.AbstractModel):
+    _inherit = "mail.activity.mixin"
+    activity_ids = fields.One2many(
+        domain=lambda self: [("res_model", "=", self._name), ("active", "=", True)]
+    )
+
+    def read_progress_bar(self, domain, group_by, progress_bar):
```
hbrunn, line 86: given upstream has a test for the active flag in [read_group](https://github.com/OCA/OCB/blob/18.0/addons/mail/models/mail_activity_mixin.py#L276), I doubt we need this override. Did you test this?

hbrunn (changes requested):

mail_activity_done/models/mail_activity.py
StefanRijnhart: Without this override, the related test fails:
```
2025-02-12 13:44:54,172 7423 ERROR 18odoo odoo.addons.mail_activity_done.tests.test_mail_activity_done: FAIL: TestMailActivityDoneMethods.test_read_progress_bar
Traceback (most recent call last):
  File "/home/bruno/odoo/18.0/mail/mail_activity_done/tests/test_mail_activity_done.py", line 60, in test_read_progress_bar
    self.assertEqual(len(result), 1)
AssertionError: 38 != 1
```

mail_activity_done/models/res_users.py
StefanRijnhart: Yes, this one can go. I'm superseding this one.

mail_activity_done/models/mail_activity.py
StefanRijnhart: Correction, the current implementation seems to raise IndexError and is harmful. The 38 records represent all the records that satisfy the filter in the call and are mapped to activity counts. Without the override, the counts are zero for done activities so the override can be removed.

StefanRijnhart:
Superseded by https://github.com/OCA/mail/pull/31

PR #23, opened 2024-12-31
[18.0][MIG] mail_attach_existing_attachment: Migration to 18.0

fredericgrall:


baf-adhoc:
HI! Is there any update on this request?

baf-adhoc:
Functional review ok

baf-adhoc (approved):
Functional review ok

david-banon-tecnativa (changes requested):
This needs some more updating, as functionality is missing on some mail composing wizards (ex Invoices)

pedrobaeza:
The sending wizard for invoices is special, and that's why it requires an extension: https://github.com/OCA/social/tree/16.0/mail_attach_existing_attachment_account

david-banon-tecnativa:
> The sending wizard for invoices is special, and that's why it requires an extension: https://github.com/OCA/social/tree/16.0/mail_attach_existing_attachment_account

I don't see why are they separate modules, however that makes this migration work as expected.

david-banon-tecnativa (approved):
Functional Review OK

pedrobaeza:
It's because you may not have `account` installed on your instance (for example, CRM only instances), so requiring it just for having the feature is too much.

/ocabot merge nobump

PR #31, opened 2025-02-12
[18.0][MIG] mail_activity_done: Migration to 18.0

StefanRijnhart:
Continued from https://github.com/OCA/mail/pull/17

The functionality is now supported by Odoo, using a similar datamodel: to mark activities done, they are archived instead of unlinked. This is reflected in the way that the state field is computed.

We can keep this module to ensure that all activity types default to keeping done activities (which is not the case in Odoo). Otherwise, the migration script will also clear archived activities that were not done, to prevent them from showing up as 'done' in Odoo 18.


hbrunn (approved):

Bart-dh:
Isn't this default functionality in Odoo 18?
image

StefanRijnhart:
@Bart-dh Ooooh, nice find! I'll check if there is remaining functionality to keep, and if migration is required for done activities from earlier versions.

StefanRijnhart:
@Bart-dh Thanks again! Indeed, Odoo 18 supports keeping done activities natively. I refactored the module to smoothen out the migration, and to ensure that the activity types in Odoo are always configured to keep done activities.

cc @hbrunn

mail_activity_done/migrations/18.0.1.0.0/post-migration.py
```diff
@@ -0,0 +1,11 @@
+from odoo.addons.mail_activity_done.hooks import _set_keep_done
+
+
+def migrate(cr, version):
+    """Set keep_done on all existing activity types"""
+    _set_keep_done(cr)
+    cr.execute(
+        """
+        delete from mail_activity where active is not true and done is not true;
```
hbrunn: this one is unexpected for me, what's your rationale here?

StefanRijnhart: All archived activites are considered done in Odoo 18: https://github.com/odoo/odoo/blob/44e61c6/addons/mail/models/mail_activity.py#L154

hbrunn: yes but we're migrating the module that people install to keep activities, so I'd not expect its migration to delete anything. I'd expect the migration script to set active to False where done is True (even though the v17 code of this module should have set this in the first place)

StefanRijnhart: True, handling done activities that are still active is another inconsistency that might need handling. As for the other case, I downgraded that to a warning.

Ricardoalso (approved):

PR #33, opened 2025-02-13
[18.0][MIG] mail_send_confirmation: Migration to 18.0

kevinkhao:
https://github.com/OCA/social/pull/1572


yostashiro (approved):
Passing along the approval on the other PR.

@kevinkhao Thanks for the migration! I think you can use the same title as the other PR. At least the module name should be corrected.

joepsanders:
@pedrobaeza is this one ready for merging? Would love to use it :)

PR #45, opened 2025-04-17
[18.0][FIX] mail_outbound_static: Use the function formataddr() from odoo.tools instead of email.utils.

BT-fgarbely:
Until now, the function `formataddr()` from `email.utils` has been used.
This leads to a problem if the sender's name contains at least one comma and at least one special character (e.g. `ü`).

Simplified example:
`formataddr('Schmid, Jürgen', 'sch.j@example.com')` results in `Schmid, Jürgen <sch.j@example.com>`.
That's wrong, because the mail servers interpret it as multiple addresses. 

Now, we can use the function `formataddr()` from `odoo.tools`.

The same example:
`formataddr('Schmid, Jürgen', 'sch.j@example.com')` results in `"Schmid, Jürgen" <sch.j@example.com>`.
That's correct.

### DIFF
--- a/mail_outbound_static/models/ir_mail_server.py
+++ b/mail_outbound_static/models/ir_mail_server.py
@@ -2,10 +2,11 @@
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
 
 import re
-from email.utils import formataddr, parseaddr
+from email.utils import parseaddr
 
 from odoo import _, api, fields, models, tools
 from odoo.exceptions import ValidationError
+from odoo.tools import formataddr
 
 
 class IrMailServer(models.Model):
@@ -84,7 +85,7 @@ class IrMailServer(models.Model):
         ):
             email_from = formataddr((name_from, mail_server.smtp_from))
             message.replace_header("From", email_from)
-            smtp_from = email_from
+            smtp_from = mail_server.smtp_from
             if not self._get_default_bounce_address():
                 # then, bounce handling is disabled and we want
                 # Return-Path = From
--- a/mail_outbound_static/__manifest__.py
+++ b/mail_outbound_static/__manifest__.py
@@ -7,7 +7,7 @@
     "version": "18.0.1.0.0",
     "category": "Discuss",
     "website": "https://github.com/OCA/mail",
-    "author": "brain-tec AG, LasLabs, Adhoc SA, Odoo Community Association (OCA)",
+    "author": "braintec AG, LasLabs, Adhoc SA, Odoo Community Association (OCA)",
     "license": "LGPL-3",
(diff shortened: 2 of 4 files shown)


BT-fgarbely:
Hi @pedrobaeza and @hbrunn 
Do you see my change as valid in functional and technical terms?

Best regards

pedrobaeza:
I don't use this module, sorry.

mail_outbound_static/models/ir_mail_server.py
```diff
@@ -84,7 +85,7 @@ def _prepare_email_message(self, message, smtp_session):
         ):
             email_from = formataddr((name_from, mail_server.smtp_from))
             message.replace_header("From", email_from)
-            smtp_from = email_from
+            smtp_from = mail_server.smtp_from
```
hbrunn, line 88: why this?

BT-fgarbely: Hi @hbrunn 
Thank you for taking a look at my PR.

By changing the use of the `formataddr()` function from `odoo.tools` instead of `email.utils`, the sender address is encoded in `utf-8`.

Example:
`"Schmid, Jürgen" <sch.j@example.com>`

This is the value that is stored in the variable `email_from`.

This means we can have accented letters here.
The `smtp_from` variable (email used during the authentication to the mail server) may only contain `ASCII` characters.

This is also ensured in the Odoo Core:
https://github.com/odoo/odoo/blob/44e61c6b429aab94e4656ade5e0f0275b905210a/odoo/addons/base/models/ir_mail_server.py#L682-L690

Since the `smtp_from` address is only used to construct the message envelope used by the transport agents and has no influence on the message header, I have found this customization to use the defined `mail_server.smtp_from` to be safe (https://docs.python.org/3/library/smtplib.html#smtplib.SMTP.sendmail).
So we only have `ASCII` characters in `smtp_from`.

If we have accented letters in `smtp_from`, certain mail providers may refuse to receive the e-mail.

Example:
Reply from **web.de**

```
** Message blocked ** Your message to [...]@web.de has been blocked. See technical details below for more information. Learn more here: https://postmaster.web.de/en/case?c=r0710 [...] (Warning: This link will take you to a third-party site) The response from the remote server was: 554 Transaction failed Reject due to policy restrictions. For explanation visit https://postmaster.web.de/en/case?c=r0710 [...]
```
https://postmaster.web.de/en/case?c=r0710

hbrunn (approved):

BT-ssteiner (approved):

hbrunn:
@BT-fgarbely btw I don't see any reviews on OCA repos done by you, you should review other people's work if you expect yours to be reviewed

PR #35, opened 2025-02-14
[18.0][MIG] base_search_mail_content : Migration to 18.0

ellbristow:
Migration to 18.0 - Previously in oca/social

Test fixed, otherwise unchanged code

App installs, unit tests pass

PR #50, opened 2025-05-26
[18.0][MIG] base_search_mail_content: Migration to version 18.0

david-banon-tecnativa:
@Tecnativa @pedrobaeza @carlos-lopez-tecnativa
TT54132


carlos-lopez-tecnativa (commented):
I think this PR (https://github.com/OCA/social/pull/1627) must be included for this module to work properly. What do you think, @pedrobaeza?

Supersedes https://github.com/OCA/mail/pull/35

pedrobaeza:
I think you haven't written on the original PR to state the problems.

> I think this PR (https://github.com/OCA/social/pull/1627) must be included for this module to work properly.

Do we gain performance with that?

base_search_mail_content/models/mail_thread.py
```diff
@@ -0,0 +1,60 @@
...
+            ("reply_to", operator, value),
+        ]
+        recs = self.env["mail.message"].search(model_domain)
+        return [("id", "in", recs.mapped("res_id"))]
+
+    message_content = fields.Text(
+        help="Message content, to be used only in searches",
+        compute="_compute_message_content",
+        search="_search_message_content",
+    )
```
carlos-lopez-tecnativa, line 34: I got the same error as in the test, but I don’t understand why. The field exists but is not stored,could a recent change in the framework have affected the behavior? Or is the migration in https://github.com/OCA/server-tools/pull/3272 incomplete? I suspect it might be due to these patches: https://github.com/OCA/server-tools/blob/18.0/base_search_fuzzy/hooks.py. Please check.
image

base_search_mail_content/tests/test_base_search_mail_content.py
```diff
@@ -0,0 +1,35 @@
+# Copyright 2017 ForgeFlow S.L.
+#   (http://www.forgeflow.com)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from odoo.tests.common import TransactionCase
+
+
+class TestBaseSearchMailContent(TransactionCase):
+    def setUp(self):
```
carlos-lopez-tecnativa: I think this can be done in the `setUpClass` method instead.

carlos-lopez-tecnativa (changes requested):

base_search_mail_content/tests/test_base_search_mail_content.py
david-banon-tecnativa: Missed that,test class is now updated to BaseCommon as well, thank you

base_search_mail_content/models/mail_thread.py
david-banon-tecnativa: The dependency for base_search_fuzzy got removed, so the operator wasn't available, fixed, thank you!

carlos-lopez-tecnativa (approved):

base_search_mail_content/__manifest__.py
```diff
@@ -0,0 +1,16 @@
...
+#   (<http://www.serpentcs.com>)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+{
+    "name": "Base Search Mail Content",
+    "version": "18.0.1.0.0",
+    "author": "ForgeFlow, SerpentCS, Tecnativa, Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/mail",
+    "category": "Social",
+    "depends": ["mail", "base_search_fuzzy"],
```
pedrobaeza, line 13: If you reintroduce the dependency, why are you not replacing the operator? And is this really needed to have that dependency/operator? Nobody has missed the fuzzy search.

david-banon-tecnativa: The operator is added by base_search_fuzzy, and i believe this module is not very useful without it, as you need to input all words fully in order. With the fuzzy dependency you can just write words, unordered ex: "fish sale kilo" would find the message: "This is about the sale of 5000 kilograms of fish" but, without it it would not.

pedrobaeza: But the module is not `base_search_fuzzy_mail_content`, but without fuzzy, and yes, it's useful, as it allows to search messages from the parent document. The problem with fuzzy is that it adds even more performance penalty. Check it on our DB with hundred thousands messages, and you will see. I think it's better to leave it without the operator nor the dependency.

carlos-lopez-tecnativa: > But the module is not `base_search_fuzzy_mail_content`, but without fuzzy, and yes, it's useful, as it allows to search messages from the parent document. The problem with fuzzy is that it adds even more performance penalty. Check it on our DB with hundred thousands messages, and you will see. I think it's better to leave it without the operator nor the dependency.

I agree with @david-banon-tecnativa. Without `base_search_fuzzy`, this module is functionally equivalent to the other one [mail_message_search](https://github.com/OCA/mail/tree/18.0/mail_message_search) and actually, this new module is better.

If you go down this path, I think it's better not to migrate this module and instead mark it as merged with `mail_message_search`

pedrobaeza: OK, I didn't remember the other module. Let's merge this one, although we will change to the other module in our instance.

pedrobaeza (approved):
/ocabot merge nobump

PR #54, opened 2025-06-15
[18.0][FIX] mail_composer_cc_bcc: emails list

norlinhenrik:
image

`mail.mail.email_to` had comma between each character. Not so anymore.

image

### DIFF
--- a/mail_composer_cc_bcc/models/mail_mail.py
+++ b/mail_composer_cc_bcc/models/mail_mail.py
@@ -8,12 +8,15 @@ from odoo.addons.base.models.ir_mail_server import extract_rfc2822_addresses
 
 
 def format_emails(partners):
-    emails = [tools.formataddr((p.name or "", p.email)) for p in partners if p.email]
-    return ", ".join(emails)
+    return [tools.formataddr((p.name or "", p.email)) for p in partners if p.email]
 
 
 def format_emails_raw(partners):
-    emails = [p.email for p in partners if p.email]
+    return [p.email for p in partners if p.email]
+
+
+def format_emails_str(partners):
+    emails = format_emails(partners)
     return ", ".join(emails)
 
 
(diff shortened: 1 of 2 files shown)


norlinhenrik:
@trisdoan @pedrobaeza 

Default behavior in `mail` is that `_prepare_outgoing_list()` returns mails with `email_to` as a **LIST**.

Currently, with mail_composer_cc_bcc `_prepare_outgoing_list()` returns `email_to` as a **STRING**. This also works somehow.

[But `mail_tracking` converts `email_to` from LIST to STRING using COMMASPACE.](https://github.com/OCA/mail/blob/18.0/mail_tracking/models/mail_mail.py#L19)

So `email_to` needs to be a list, to avoid the error described above.

pedrobaeza:
I don't use this module, sorry.

trisdoan:
Hello @norlinhenrik, you are right, thanks for the fix, but CI is red, could you take a look please?

Also, I got error: Invalid email address

norlinhenrik:
I have no idea why a test is failing here. It passes locally.
image

mail_composer_cc_bcc/tests/test_mail_cc_bcc.py
```diff
@@ -84,15 +84,13 @@ def test_email_cc_bcc(self):
...
-                '"Marc Demo" <mark.brown23@example.com>',
-                '"Joel Willis" <joel.willis63@example.com>',
-                '"Chester Reed" <chester.reed79@example.com>',
-            ]
-        )
+        expecting = [
+            '"Marc Demo" <mark.brown23@example.com>',
+            '"Joel Willis" <joel.willis63@example.com>',
+            '"Chester Reed" <chester.reed79@example.com>',
+        ]
```
trisdoan: I don't think it's correct, it should be string

trisdoan (changes requested):

BT-anieto:
@norlinhenrik Is there any advancements here?

norlinhenrik:
@trisdoan @BT-anieto
Can you please test this PR, and approve it if it works as expected?
I believe it is ready to merge.

norlinhenrik:
@trisdoan @BT-anieto
Finally the tests are passing.
Can you please review?

trisdoan (approved):

ChristophAbenthungCibex (approved):
LGTM :+1:

PR #58, opened 2025-06-30
[18.0] mail_tracking: Apply read restrictions to everyone not being superuser

grindtildeath:
The previous condition to apply access restriction to mail tracking email records was not applied to every user having access rights settings. Hence, this could lead up to AccessError when trying to open the model's action through the web client.

Since the access to these records is based on the same restrictions that are implemented for mail.message model in the mail module, we ended up having access error as there were records we were allowed to search but not to read, due to the restriction being more permissive in the case of mail tracking email.

But since the check_access method of this model is actually checking if the user has access to the related mail.message, we need to have the same conditions to avoid such access error for users with settings access that are not superuser.

### DIFF
--- a/mail_tracking/models/mail_tracking_email.py
+++ b/mail_tracking/models/mail_tracking_email.py
@@ -152,7 +152,7 @@ class MailTrackingEmail(models.Model):
         _check_access() for more details about those rules.
         """
         query = super()._search(domain, offset, limit, order)
-        if not self.env.user.has_group("base.group_system"):
+        if not self.env.is_superuser():
             records = self.browse(query)
             allowed_ids = self._get_allowed_ids(records.ids)
             return self.browse(allowed_ids)._as_query(order)


grindtildeath:
ping @pedrobaeza @chienandalu @hbrunn :pray:

twalter-c2c (approved):

chienandalu (approved):
LGTM :+1:

pedrobaeza:
@grindtildeath please change commit message to prefix `[FIX]` tag like commit message guidelines say: https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#commit-message

mmequignon (approved):
Just the tag to add in the commit, but looks good, thanks!

grindtildeath:
@pedrobaeza @mmequignon Commit message amended

pedrobaeza (approved):
/ocabot merge patch

sebalix:
@grindtildeath Is this patch also relevant for 17.0 and 16.0?

grindtildeath:
@sebalix I'm not sure, but I guess it could be needed from the versions where we had the split between odoobot and administrator

PR #59, opened 2025-07-02
[ADD] mail_user_signature_template: Company-wide signature templates

jwaes:
This module provides company-wide email signature templates that automatically populate with user information, ensuring consistent and professional email signatures across the organization.

Features:
- Create and manage multiple signature templates per company
- Automatic population with user data (name, email, phone, function, etc.)
- Company-level configuration with optional enforcement
- Template preview functionality
- Multi-company support
- Comprehensive security model with user/manager groups
- Support for company logos with positioning options

Technical implementation:
- Extends res.users to override signature computation
- Adds signature template configuration to res.company
- Uses mail.render.mixin for template rendering


ChristophAbenthungCibex:
Hi @jwaes, 
thanks for this module. Overall it looks pretty decent.

Some errors i found during my tests:
- When disabling company logo at signature_template form view, its not possible to enable it again because the field is in the group that gets invisible if company logo is disabled
- Unclear if mail is with signature or without (in chatter)
	1. When sending without composer -> no signature in chatter but signature automatically added to mail
	2. When sending with composer -> signature added in composer and chatter shows signature
	3. When sending with composer and removing signature from mail -> not added again and chatter shows no signature -> same look as case 1 -> mail for case 1 has the signature added
- When company enforces signature -> go to user -> preferences -> try to change signature template -> an error is thrown
- Changes in users phone number and email don't seem to change for template (not even after logout and login)
- The method _render_signature for signatures prevents users from being created
- When user is in multiple companies -> user changes to second company (non main company and only this one is selected) -> template of main company is used instead of current company (shows incorrect company name too)

PR #65, opened 2025-07-08
[18.0][MIG] mail_attach_existing_attachment_account: Migration to 18.0

ByteMeAsap:


carlos-lopez-tecnativa (approved):
LGTM

mail_attach_existing_attachment_account/tests/test_mail_attach_existing_attachment_account.py
```diff
@@ -9,10 +9,10 @@ class TestMailAttachExistingAttachmentAccount(BaseCommon):
     @classmethod
     def setUpClass(cls):
         super().setUpClass()
-        cls.product = cls.env["product.product"].create({"name": "Test product"})
-        cls.partner = cls.env["res.partner"].create({"name": "Mr. Odoo"})
+        cls.product = cls.env["product.product"].create([{"name": "Test product"}])
```
pedrobaeza: No need of these changes. It works the same.

mail_attach_existing_attachment_account/wizard/account_move_send_wizard.py
```diff
@@ -20,28 +20,20 @@ class AccountMoveSend(models.TransientModel):
         compute="_compute_display_object_attachment_ids",
     )
 
-    @api.depends("move_ids")
+    @api.depends("move_id")
```
pedrobaeza, line 23: The wizard should work the same for several invoices, isn't it?

pedrobaeza (commented):
@carlos-lopez-tecnativa does this have the same problem with Odoo core as in 17?

mail_attach_existing_attachment_account/wizard/account_move_send_wizard.py
ByteMeAsap: There is a separate wizard if you want to send it for several invoices in v18. Also, this module was always working for a single invoice only.

mail_attach_existing_attachment_account/tests/test_mail_attach_existing_attachment_account.py
ByteMeAsap: Reverted the changes

carlos-lopez-tecnativa:
> @carlos-lopez-tecnativa does this have the same problem with Odoo core as in 17?

Yes, the problem exists in V18 too, and it must be handled by https://github.com/odoo/odoo/pull/215420

ByteMeAsap:
@pedrobaeza , what is the next step/action needed for this?

pedrobaeza:
AFAIK, this doesn't work unless Odoo accepts the PR.

ByteMeAsap:
> AFAIK, this doesn't work unless Odoo accepts the PR.

@pedrobaeza , the PR is accepted, so this should now be mergeable

pedrobaeza (approved):
/ocabot merge nobump

PR #68, opened 2025-07-21
[18.0][REM] mail_debrand: Moved to OCA/server-brand

MiquelRForgeFlow:
We move all the brand/debrand modules to a specific repo for better organization.

- [ ] https://github.com/OCA/server-brand/pull/101.


pedrobaeza:
The attribution should be preserved in the migration to 18 in OCA/server-brand, but the rest is OK.

MiquelRForgeFlow:
> The attribution should be preserved in the migration to 18 in OCA/server-brand, but the rest is OK.

Done.

JordiBForgeFlow:
@pedrobaeza can you close it, as agreed?

PR #72, opened 2025-08-12
[FIX] mail_message_search : prevent db crach

fmdl:
Before this commit a search in large database can crash, because the return of search() can include million of ids.

This PR use SQL (and not python).
A limit can be set to prevent a crash.

### DIFF
--- a/mail_message_search/models/mail_thread.py
+++ b/mail_message_search/models/mail_thread.py
@@ -26,8 +26,10 @@ class MailThread(models.AbstractModel):
                 word_domain_list.append(expression.OR(field_domain_list))
         word_domain = expression.AND(word_domain_list)
         domain = expression.AND([[("model", "=", self._name)], word_domain])
-        messages = self.env["mail.message"].search(domain)
-        return [("id", "in", messages.mapped("res_id"))]
+        limit_value = self.env['ir.config_parameter'].sudo().get_param("mail.message.search.limit")
+        limit = int(limit_value) if limit_value else None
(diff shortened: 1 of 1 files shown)


fmdl:
@alexis-via

mail_message_search/models/mail_thread.py
```diff
@@ -26,8 +26,10 @@ def _search_message_search(self, operator, value):
                 word_domain_list.append(expression.OR(field_domain_list))
         word_domain = expression.AND(word_domain_list)
         domain = expression.AND([[("model", "=", self._name)], word_domain])
-        messages = self.env["mail.message"].search(domain)
-        return [("id", "in", messages.mapped("res_id"))]
+        limit_value = self.env['ir.config_parameter'].sudo().get_param("mail.message.search.limit")
```
yostashiro, line 29: Something like this may follow the convention better.
```suggestion
        limit_value = self.env['ir.config_parameter'].sudo().get_param("mail_message_search.message_limit")
```

Can you please add the CONFIGURE file for this?

yostashiro (commented):
Thanks for the fix!

AungKoKoLin1997:
@fmdl Thanks for the fix. I backported this PR in my 17.0 migration PR and also added configuration file.
It’d be great if you could include the configuration file in your PR. Please also fix the pre-commit error.

AungKoKoLin1997:
Forgot to attach the backported commit in my above commend.
https://github.com/OCA/social/pull/1704/commits/6f374ca0305dfc8d312aff23b3fb310d20e8a9c1

PR #73, opened 2025-09-01
[18.0][IMP]mail_quoted_reply: add reply to all

ChristophAbenthungCibex:
Add a reply to all button to messages. Default partners is the sender and all recipients of the original message without the user sending it.

Based on PR https://github.com/OCA/social/pull/1688


OKT96 (approved):
LGTM :)

PR #74, opened 2025-09-02
[18.0][FIX]mail_quoted_reply: unintentional tags

ChristophAbenthungCibex:
<...> text of subject and from in the quoted reply block are interpreted as tags and thus not shown in the composer. The unintentional tag may lead to browser problems because the browser sees it as an opening tag without a closing one.

### Steps to reproduce
Reply to any message. The composer will open and from field at the reply will only show the name and not the e-mail address. Inspecting with the developer tools of the browser shows that the e-mail address is interpreted as a tag. The same behavior can be observed for the subject for the reply (in the quoted reply block)

Port of  https://github.com/OCA/social/pull/1435

### DIFF
--- a/mail_quoted_reply/models/mail_message.py
+++ b/mail_quoted_reply/models/mail_message.py
@@ -2,7 +2,7 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 from odoo import _, models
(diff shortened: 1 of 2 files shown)


dhx (approved):
Looks good!

PR #78, opened 2025-09-08
[18.0] [IMP] `mail_tracking`: remove redundant ui FailedMessagePanel

ivantodorovich:
There's already a collapsable panel for failed messages in the chatter.
Having this other panel is totally redundant, and it was a bit glitchy too, probably not fully migrated to 18.0.. as it was working a little bit better in 17 (although, also redundant IMO)

https://github.com/user-attachments/assets/3e6e304e-4b35-45f5-9d10-703fe635b0fa


pedrobaeza (approved):
Check the before/after in runboat and indeed it's redundant.

PR #90, opened 2025-10-01
[19.0][MIG] mail_post_defer: Migration to 19.0

yp-oerp:
Standard migration to v19

@yajo, please help to review if possible


yp-oerp:
@yajo @u0f @nobuQuartile 

This is ready, please help review this PR

Thank you

yajo (changes requested):
I have tested on runboat and I cannot see any message being sent after a minute. Could you please check it and provide testing steps?

Thanks.

AungKoKoLin1997 (approved):
Functional Review: LGTM

I tested the functionality on Runboat.
Without this module, I noticed that an email appeared immediately in the Emails list with a Delivery Failed status when sending from the partner chatter.
With this module installed, the emails appear after a short delay.

yp-oerp:
@AungKoKoLin1997  I’ve been swamped and haven’t had a chance to re-test this yet..Thanks for helping with the testing!

@yajo Let me know if anything else is needed from my side before we merge this

Thank you!

mail_post_defer/models/mail_thread.py
```diff
@@ -0,0 +1,76 @@
...
+            kwargs.setdefault("force_send", force_send)
+            if not force_send:
+                # If deferring message, give the user some minimal time to revert it
+                _self = _self.with_context(mail_defer_seconds=30)
+        # Apply deferring
+        defer_seconds = _self.env.context.get("mail_defer_seconds")
+        if defer_seconds:
+            kwargs.setdefault(
+                "scheduled_date",
+                datetime.utcnow() + timedelta(seconds=defer_seconds),
```
yostashiro: To be safe, since `datetime.utcnow()` is deprecated in Python 3.12.
```suggestion
                fields.datetime.now() + timedelta(seconds=defer_seconds),
```

mail_post_defer/models/mail_thread.py
```diff
@@ -0,0 +1,76 @@
...
+            raise UserError(
+                self.env._("Cannot modify message; notifications were already sent.")
+            ) from None
+
+        # For unsent messages, we're more permissive than the original implementation
+        # Only check the basic restrictions that make sense
+        if any(message.message_type != "comment" for message in messages):
+            raise UserError(
+                self.env._("Only messages type comment can have their content updated")
+            )
```
yostashiro: I think this should be dropped as the extra checks block updates to sent comments, which are allowed in the standard functionality. The equivalent checks can be implemented in `_message_update_content()` instead.
```suggestion
```

mail_post_defer/models/mail_thread.py
```diff
@@ -0,0 +1,76 @@
...
+        res = super()._message_update_content(message, body=body, **kwargs)
+        # Delete empty pending outgoing mails
+        if empty_messages := message.sudo()._filter_empty():
+            empty_messages.mail_ids.filtered(
+                lambda mail: mail.state == "outgoing"
+            ).unlink()
+            empty_messages.env["mail.message.schedule"].search(
+                [("mail_message_id", "in", empty_messages.ids)]
+            ).unlink()
+        return res
```
yostashiro: I noticed an issue of mail.mail records not being created when the message is edited before sent. The following code makes sense to me. Can you please check?
```suggestion
        # If anything already went out, fall back to the standard behavior
        if any(ntf.notification_status == "sent" for ntf in message.notification_ids):
            return super()._message_update_content(message, body=body, **kwargs)
        scheduled_date = fields.Datetime.now() + timedelta(seconds=30)
        Schedule = self.env["mail.message.schedule"].sudo()
        sched = Schedule.search([("mail_message_id", "=", message.id)], limit=1)
        if sched:
            sched.scheduled_datetime = scheduled_date
        # In case emails are already created but not sent yet, cancel queued mails and
        # delete the notification (so nothing goes out with old content)
        else:
            message.mail_ids.filtered(
                lambda x: x.state in ("outgoing", "exception", "draft")
            ).write({"state": "cancel"})
            message.notification_ids.filtered(
                lambda x: x.notification_status in ("ready", "exception", "canceled", "bounce")
            ).unlink()
            Schedule.create(
                {"mail_message_id": message.id, "scheduled_datetime": scheduled_date}
            )
        kw = dict(kwargs)
        # Drop empty [] for partner_ids, or else Odoo will try to send to nobody
        if not kw.get("partner_ids"):
            kw.pop("partner_ids", None)
        return super()._message_update_content(message, body=body, **kw)
```

mail_post_defer/readme/CONFIGURE.md
```diff
@@ -0,0 +1,19 @@
+You usually don't need to do anything. The module is configured appropriately out of the box.
+Just make sure the following scheduled actions are active:
+
+- Mail: Email Queue Manager (mail.ir_cron_mail_scheduler_action)
+- Notification: Send scheduled message notifications (mail.ir_cron_send_scheduled_message)
```
yostashiro:
```suggestion
- Notification: Notify scheduled messages (mail.ir_cron_send_scheduled_message)
```

yostashiro (changes requested):

u0f (changes requested):
I've been testing on Runboat and I can't see any message send after the time. Could you check it?

yp-oerp:
@yajo @u0f 

I also tested in the OCA runboat, it seems to work fine. Could you please let me know how you are testing it? For your reference I use this contact to test "SONIC HEALTHCARE LIMITED"

I see @AungKoKoLin1997 was also able to confirm the same.

AungKoKoLin1997:
> I see @AungKoKoLin1997 was also able to confirm the same.

@yp-oerp In my last review, I only tested creating a new email, and it worked fine.
Today, I also tested editing the email during the defer period, and it turned out it wasn’t sent, as @yostashiro mentioned.
https://github.com/OCA/mail/pull/90#discussion_r2483788363

I didn’t see the email icon when I edited the message.
Screenshot 2025-11-07 084950

This is the message from the unedited email.
Screenshot 2025-11-07 085206

This is the message from the edited email.
Screenshot 2025-11-07 085258

mail_post_defer/models/mail_thread.py
yp-oerp: It make sense, I added `fields.Datetime.now()`

mail_post_defer/models/mail_thread.py
yp-oerp: @yostashiro I made some changes, now I'm calling super and removing some extra checks that Odoo was already checking...

But the helper method is still useful because Odoo does not "resent" the message after it's edited, but with this module, there is a danger that if we allow sending "edited sent" message, the end receiver will receive two copies of the message. So it makes sense to have the first two checks.

mail_post_defer/models/mail_thread.py
yp-oerp: @yostashiro thank you for the suggestion the main issue is that Odoo base is passing empty "partner_ids" and that was removing the partners(thus, no email sent)...I added the check about the "partner_ids" (from your suggestion) to my original code and it seems to work fine..

As per my above comment, I have kept the code as originally implemented

yp-oerp:
@AungKoKoLin1997 @yostashiro @yajo @u0f thank you for your help with code review and testing.

I made some changes, and I tested them in [Runboat](http://oca-mail-19-0-pr90-a1d87eb50261.runboat.odoo-community.org/); it seems to work fine. Please feel free to review further and let me know

Screenshot 2025-11-07 at 12 13 41 PM
 

Thank you

mail_post_defer/models/mail_thread.py
```diff
@@ -0,0 +1,74 @@
...
+            kwargs.pop("partner_ids", None)
+        res = super()._message_update_content(message, body=body, **kwargs)
+        # Delete empty pending outgoing mails
+        if empty_messages := message.sudo()._filter_empty():
+            empty_messages.mail_ids.filtered(
+                lambda mail: mail.state == "outgoing"
+            ).unlink()
+            empty_messages.env["mail.message.schedule"].search(
+                [("mail_message_id", "in", empty_messages.ids)]
+            ).unlink()
```
yostashiro, line 69: I think this part is fine for the scope of the comment, but it doesn't cover the case where the message content is changed (not emptied) after the mail is created but before it's sent. In that situation, the email would still be sent with the old content. Can you please review once again the logic I suggested in https://github.com/OCA/mail/pull/90#discussion_r2483788363?

yostashiro (changes requested):

mail_post_defer/models/mail_thread.py
yostashiro: @yp-oerp I think it's a matter of preference, and whether these checks are applied or not should ideally be made configurable. In any case, it would be nice to have a comment in the README on this since it changes the standard behavior.

mail_post_defer/models/mail_thread.py
yp-oerp: @yostashiro I have incorporated the changes you suggested. I was able to verify the issue and fix it with your suggestion.

I had to kept the empty messages part (`empty_messages := message.sudo()._filter_empty()`) as it was trying to send

yp-oerp:
@yostashiro changes are pushed, please help to review and test

Thank you

yostashiro (approved):
Code review and functional test. LTGM. 👍 

Thanks for the update!

mail_post_defer/models/mail_thread.py
```diff
@@ -27,41 +26,45 @@ def _notify_thread(self, message, msg_vals=False, **kwargs):
...
+        scheduled_date = fields.Datetime.now() + timedelta(seconds=30)
+        Schedule = self.env["mail.message.schedule"].sudo()
+        sched = Schedule.search([("mail_message_id", "=", message.id)], limit=1)
+        if sched:
+            sched.scheduled_datetime = scheduled_date
+        # In case emails are already created but not sent yet, cancel queued mails and
+        # delete the notification (so nothing goes out with old content)
+        else:
+            message.mail_ids.filtered(
+                lambda x: x.state in ("outgoing", "exception", "draft")
```
yajo: suggestion: better performance

```suggestion
                lambda x: x.state in {"outgoing", "exception", "draft"}
```

mail_post_defer/models/mail_thread.py
```diff
@@ -27,41 +26,45 @@ def _notify_thread(self, message, msg_vals=False, **kwargs):
...
+            ).write({"state": "cancel"})
+            message.notification_ids.filtered(
+                lambda x: x.notification_status
+                in ("ready", "exception", "canceled", "bounce")
```
yajo:
```suggestion
                in {"ready", "exception", "canceled", "bounce"}
```

mail_post_defer/readme/CONFIGURE.md
```diff
@@ -0,0 +1,19 @@
...
+this module configures that job to execute every minute by default.
+
+You can still change that cadence after installing the module (although
+it is not recommended). To do so:
+
+1.  Log in with an administrator user.
+2.  Activate developer mode.
+3.  Go to *Settings \> Technical \> Automation \> Scheduled Actions*.
+4.  Edit the action named "Mail: Email Queue Manager".
+5.  Lower down the frequency in the field *Execute Every*. Recommended: 1 minute.
```
yajo: issue(ux): I think that this should happen too with the `mail.ir_cron_send_scheduled_message` cron job. Otherwise, that one is done each 1h out of the box, and also impacts UX negatively.

These docs should be updated too.

yajo (changes requested):
Functional tests were OK, thanks for the updates.

I suggest a couple of minor performance enhancements, and a more important UX improvement.

mail_post_defer/readme/CONFIGURE.md
yp-oerp: I updated the hook to target the `mail.ir_cron_send_scheduled_message` as well and updated docs too

yp-oerp:
@yajo I improved the module and added the improvements that you suggested

Please help to review

Thank you

yajo (approved):

yp-oerp:
@u0f Please help to review if possible (It seems it needs one more review to merge)

yp-oerp:
@yostashiro @AungKoKoLin1997 @u0f 

Wondering if anyone can help with the second review?

u0f (approved):
LGTM! Functional Review done in Runboat.

PR #103, opened 2025-10-29
[18.0][FIX] mail_message_search: Error a bytes-like object is required, not str

AungKoKoLin1997:
This module causes the same issue with https://github.com/OCA/mail/pull/102#issue-3562387721 if we don't have `encoding`.

@qrtl QT5667


yostashiro (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #107, opened 2025-11-06
[19.0][MIG] mail_outbound_static

astirpe:


CasVissers-360ERP (approved):
Functional review

CasVissers-360ERP:
Actually there is an issue with the from e-mail address with this version. The name disappears in the from (it's only the e-mail address).

astirpe:
@CasVissers-360ERP Thanks! Should be fixed now, could you check again?

CasVissers-360ERP:
@astirpe all good now! Thanks!

mail_outbound_static/models/ir_mail_server.py
```diff
@@ -32,8 +32,9 @@ def check_valid_domain_whitelist(self):
             domains = list(self.domain_whitelist.split(","))
             for domain in domains:
                 if not self._is_valid_domain(domain):
+                    # pylint: disable=translation-not-lazy
```
StefanRijnhart: Is this not easily fixed by rewriting the error message as `self..env._("%(domain)s is not ...", domain=domain)`?

astirpe: @StefanRijnhart  Fixed, thanks!

StefanRijnhart (approved):
Thanks!

astirpe:
@CasVissers-360ERP @StefanRijnhart 
I added a necessary change, fixing a bug discovered after using this module in production. I temporary kept it in a separate commit for easier review.

CasVissers-360ERP:
I can confirm that issue is fixed now.

StefanRijnhart (approved):
Thanks for the update! If the change would not apply to earlier version of Odoo it would be best to squash the commits.

astirpe:
Squashed and rebased, thanks!

StefanRijnhart (approved):

Reyes4711-S73 (approved):
LGTM

PR #110, opened 2025-11-07
fix _search recursion and access_rights handling on Odoo 18

thom2023:
This PR fixes multiple compatibility issues of `mail_tracking` with Odoo 18:

- Avoid passing `count` to BaseModel._search (not supported in Odoo 18)
- Apply access rights via `with_user(access_rights_uid)`
- Prevent recursion by calling the parent `_search` using MRO lookup
- Make `_get_allowed_ids` tolerant to empty/partial SQL results

Tested on:
- Odoo 18.0-20250606
- PostgreSQL 15
- Docker environment

This resolves recursion errors and restores correct mail tracking visibility behavior.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -86,3 +86,4 @@ docs/_build/
 
 # OCA rules
 !static/lib/
+*.bak.*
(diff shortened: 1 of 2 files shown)


thom2023:
This PR resolves recursion issues and fixes access rights behavior in mail_tracking on Odoo 18.
The implementation avoids passing `count` to BaseModel._search and applies user context correctly via `with_user`.
Tested on Odoo 18.0 with PostgreSQL 15.

PR #116, opened 2025-11-20
[18.0][IMP] mail_activity_team: assign plan activities to teams

StefanRijnhart:
When creating activity plans, instead of assigning an activity to a user, there is also the option to assign it to a team instead.


CRogos (commented):
Basically the change is working. 

In the normal Activity view, you can select a team_id and a team_user_id, this is missing in the plan dialog.
image

image

Not sure if we need a dedicated variable team_user_id in "mail.activity.plan.template" like in "mail.activity" or if we just should set the user_id field?

There is also room for more test coverage ;)

StefanRijnhart:
@CRogos thanks for the review! I've added a fixup-commit that improves UI consistency and test coverage. If you agree I will squash the commits.

mail_activity_team/models/mail_activity_plan_template.py
```diff
@@ -0,0 +1,95 @@
...
+    )
+    # Add compute method to existing field
+    responsible_id = fields.Many2one(
+        compute="_compute_responsible_id",
+        readonly=False,
+        store=True,
+    )
+    responsible_type = fields.Selection(
+        ondelete={"team": "set default"},
+        selection_add=[("team", "Team")],
```
CRogos, line 36: I think we should remove this option. I know selecting "Default User" is not correct named, but I think we should use it anyway and change the UI behavior the same as the create activity works.
The UI should react whether the activity_team_id is set or not. Than you can also get rid off activity_team_required.

CRogos (commented):
I agree with most of the changes, but I would change the UI behavior the same as it is here for "Default User". Maybe we can change the display name of this value to "Default User/Team"?

image

mail_activity_team/models/mail_activity_plan_template.py
StefanRijnhart: I would prefer to keep this option because `hr` adds additional types with their own display logic, and this way we keep things nice and tight (see https://github.com/odoo/odoo/blob/18.0/addons/hr/models/mail_activity_plan_template.py#L11-L15).

CRogos: Yes, but in these cases, there should nothing be selected and the field needs to be hidden. 

I still think this is not necessary, because it is different in the activity creation form. 
(offtopic) It is very sad, that the manager, coach activity is only available in the activity plan and not also available in the create activity form (strange). 

If you insist on the "team" than always show the "team user" field, event though no team was selected yet. The UI is changing to often which makes is less usable.
image

StefanRijnhart: > If you insist on the "team" than always show the "team user" field, event though no team was selected yet. The UI is changing to often which makes is less usable.

Good one, fixed now.

mail_activity_team/models/mail_activity_mixin.py
```diff
@@ -44,6 +44,13 @@ def activity_schedule(
         user-team missmatch. We can hook onto `act_values` dict as it's passed
         to the create activity method.
         """
+        # Pick up some defaults from mail.activity.schedule
+        for field in ("team_id", "team_user_id", "user_id"):
+            if self.env.context.get(f"schedule_default_{field}") and not act_values.get(
+                field
+            ):
+                act_values[field] = self.env.context[f"schedule_default_{field}"]
+
```
CRogos, line 53: Shouldn't the schedule_default_team_id be selectable here when Assignemnt is set to "ask at launch":

image

CRogos (approved):
Code + Function LGTM, but I am still not convinced by the "Assignment: Team" option, because a lot of code wouldn't be necessary if we remove it.  (e.g. activity_team_required)

I've added one more improvement suggestion.

mail_activity_team/models/mail_activity_mixin.py
StefanRijnhart: That could be a future addition, but it is out of scope for us at this point.

StefanRijnhart:
Commits squashed.

MohamedOsman7 (approved):
functional & code LGTM

CRogos:
@OCA/mail-maintainers could you review/merge?

CRogos:
@OCA/mail-maintainers could you merge?

CRogos:
@etobella could you merge this? PS: i think @OCA/mail-maintainers is not working?

MohamedOsman7:
@pedrobaeza Could you merge?

PR #117, opened 2025-11-25
[18.0][FIX] mail_tracking: Prevent error when the SQL result is empty.

carlos-lopez-tecnativa:
Before this commit, when the user clicked on the icon of a message, the following error occurred: `ValueError: not enough values to unpack (expected 4, got 0)`

Complementary to 86e188c1599638071621f850a0fe875afedd851c

image

TT59417
@Tecnativa @pedrobaeza @carlosdauden @sergio-teruel  @remi-filament could you please review this?


pedrobaeza (approved):
Fast-tracking this for fixing the error in this case:

/ocabot merge patch

PR #120, opened 2025-12-04
[18.0][IMP] mail_activity_board/mail_activity_team: Move spreadsheet_dashbord dependency to mail_activity_dashboard

CRogos:
I understand that changes of this type can be problematic, but this module is in "beta" and also need some "love".

First of all, some added groups are identical:
image

And also the native Kanban view looks a lot better than the added.
image
image

Second is there a reason for setting the mode to "primary"? When we extend the existing view, we have the same options also in the View all activities. 
image

Third, I would like to have a general improved activity view, without the need of spreadsheet_dashboard. Therefore I've moved this part to a separate spreadsheet_dashbord module. 

The current name "mail_activity_board" is not optimal, because it is more a "mail_activity_view" module, but I would like to keep the name because of migration reasons. 

@StefanRijnhart @bizzappdev Could you review and share your thoughts on this changes?


mail_activity_board/views/mail_activity_view.xml
```diff
@@ -124,119 +50,22 @@
...
             </xpath>
 
-            <xpath expr='//filter[@name="filter_date_deadline_past"]' position='before'>
+            <xpath
+                expr='//filter[@name="filter_date_deadline_future"]'
+                position='before'
+            >
                 <filter
                     string="Act. next month"
                     name="activities_month"
```
CRogos: I am not a big fan of "Act. Next month" and "Act. Next 6 Months" filter because they are very specific. Maybe we should remove them as well?

CRogos:
@atchuthan @atangaraj could you have a look at these improvements.

MohamedOsman7 (approved):
Code & Functional Review LGTM

atchuthan (approved):

mail_activity_board/migrations/18.0.1.1.0/pre-migration.py
```diff
@@ -0,0 +1,41 @@
...
+        """,
+        (module_name,),
+    )
+
+
+def rename_xmlid(cr, old, new):
+    cr.execute(
+        """
+        UPDATE ir_model_data
+        SET name = %s
```
StefanRijnhart: In my test database, the records were not migrated. You need to split up the xmlids and also select by module (see https://github.com/OCA/openupgradelib/blob/master/openupgradelib/openupgrade.py#L1347-L1350).

mail_activity_board/views/mail_activity_view.xml
StefanRijnhart: Agree! You can replace with something like

```
<filter name="date_deadline" date="date_deadline" start_year="0" start_month="-1" end_year="2" end_month="2" />
```

Image

StefanRijnhart (commented):
Thanks! Looking way better now.

atangaraj (approved):
Functional review :+1:

mail_activity_board/migrations/18.0.1.1.0/pre-migration.py
CRogos: Could you recheck the migration script?

StefanRijnhart (commented):
Migration script works fine now!

One more thing: it turns out you also need to manually bump mail_activity_team's version number. Previously it would be updated automatically because its dependency mail_activity_board is being updated, but this is now prevented because the dependency is dropped in the change.

CRogos:
OK? But there are also changes in the mail_activity_team module. Shouldn't trigger this the version update itself? But I am totally unaware of the magic behind the version update :D

I've increased the version in the manifest.

mail_activity_dashboard/readme/DESCRIPTION.md
```diff
@@ -0,0 +1 @@
+This module adds a link to the activity board in the Dashbard menu controlled by
```
StefanRijnhart: Did you mean 
```suggestion
This module adds a link to the activity board in the Dashbard menu.
```

StefanRijnhart (commented):
Thanks for updating the version number. OCA (/Acsone) tools like click-odoo-update and module_auto_update will trigger updates based on inspection of the module files themselves, but Odoo.sh's primitive mechanism will only update modules based on the version in the manifest.

StefanRijnhart (approved):
Thanks for the update, and nice work!

/ocabot merge nobump

PR #125, opened 2025-12-17
[18.0][FIX] mail_message_search : prevent db crash

AungKoKoLin1997:
Supersedes https://github.com/OCA/mail/pull/72

@qrtl QT5088


nobuQuartile (approved):
Functional review

flotho (commented):
Hi 
Thanks for the PR.
I get the point.
Yet I'm wondering why you didn't defined a default value as a system parameter.
For sure one can argue that "BIG" database depends on so many  factor as the server size, setup, etc... yet if you're pointing 5000 in you exemple maybe you've experienced such kind of limitation.
hat is you PoV on this?

Regards

AungKoKoLin1997:
> Yet I'm wondering why you didn't defined a default value as a system parameter.
> For sure one can argue that "BIG" database depends on so many factor as the server size, setup, etc... yet if you're pointing 5000 in you exemple maybe you've experienced such kind of limitation.

@flotho 
I’m not sure I fully understand your point.
My preference was to avoid setting a default limit, because the module is intended to run without any limitation in most environments. The limitation is meant to be applied only for databases with a very large volume of data.
We’ve documented this in `CONFIGURE.md:` users can set the limit when they actually need it for their environment.

flotho (approved):
LGTM

yostashiro (approved):
I think it'd be good to have a default value as @flotho suggests, but I'm not against having no value either.

etobella (approved):
Makes sense

PR #127, opened 2025-12-23
[18.0][FIX] mail_tracking: Error on tabs without thread open when marking as reviewed

CarlosRoca13:
Before these changes, when the bus event was received in tabs that did not have the message open, an error appeared because the system was not able to find the message.

After these changes, if the message is not available on the screen, the process continues because synchronization will not be necessary, and therefore the error will no longer occur.

cc @Tecnativa

ping @pedrobaeza @carlos-lopez-tecnativa


pedrobaeza (approved):
Fast-tracking, as it has been checked that it's fixing the annoying crash on the other tabs:

/ocabot merge patch

PR #129, opened 2026-01-04
[19.0][MIG] mail_activity_unlink_log: migrate to 19.0

TheSumitBasak:
Migration of `mail_activity_unlink_log` module to Odoo 19.0.

Moved from `OCA/social` to `OCA/mail` as requested.


pedrobaeza:
Thanks for the move, but you should keep the commit history. That can be done following https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-19.0, but adding the 2 remotes (social and mail), and touching slightly the command for using the social remote instead of `origin`.

PR #133, opened 2026-01-08
[18.0][FIX] mail_show_follower: handle case where record no longer ex…

kevinkhao:
…ists

In the case where the original record from the mail has been deleted, attempting to access any value on it raises:

MissingError: Record does not exist or has been deleted

As a simple solution, this fix skips the module's functionality when the record no longer exists. A more complete solution would be to personalize the message body somewhere upstream, however that would require more changes in the way the module is structured. As of now, this fix seems sufficient for exceptional cases.


yajo (approved):
/ocabot merge patch

PR #136, opened 2026-01-12
[FIX] mail_post_defer: Fixed behavior of message sending wizard

sabrinaRMartin:
Fixed the behavior of the message sending wizard so that messages can be edited after they have been sent.

@yajo @moduon MT-13187

https://www.loom.com/share/da263aabb28948218b5cd1eade79eb0e

### DIFF
--- a/mail_post_defer/__init__.py
+++ b/mail_post_defer/__init__.py
@@ -1,2 +1,3 @@
 from . import models
+from . import wizards
 from .hooks import post_init_hook
--- a/mail_post_defer/tests/test_mail.py
+++ b/mail_post_defer/tests/test_mail.py
@@ -272,6 +272,48 @@ class MessagePostCase(MailPostDeferCommon):
                 )
 
 
+@freezegun.freeze_time("2023-01-02 10:00:00")
+class ComposerCase(MailPostDeferCommon):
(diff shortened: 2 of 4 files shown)


mail_post_defer/models/mail_compose_message.py
```diff
@@ -0,0 +1,15 @@
+# Copyright 2025 Moduon Team S.L. <info@moduon.team>
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl).
+
+from odoo import models
+
+
+class MailComposer(models.TransientModel):
```
yajo: issue: this is a `TransientModel` (a wizard); therefore, it should be inside the `wizards/` folder, not inside `models/`.

yajo (changes requested):
One minor code comment.

Functionally, I found a more problematic issue: https://www.loom.com/share/6c2685041e844750932a06ad0645b00e

Gelojr (approved):
Well done!
I tried the solution suggested in the @yajo 's comment: https://github.com/OCA/mail/pull/136#pullrequestreview-3654355775, and the correct text now appears in both messages after the modification.

Image

yajo (approved):
/ocabot merge patch

PR #145, opened 2026-02-05
[18.0][IMP] mail_restrict_follower_selection: enhance domain filtering forfollower selection

CRogos:
1) I don't think that filtering on employees in default is a good setup. After installation of the module, the default behavior should not be changed.

2) There were dialogs, where the filter did not work in the search, but restricted/removed it on save. After save it looked like it was added, but after a page reload the new follower was not there. 
This occurs in the enterprise helpdesk team form.

image

### DIFF
--- a/mail_restrict_follower_selection/README.rst
+++ b/mail_restrict_follower_selection/README.rst
@@ -52,9 +52,15 @@ certain record type (or have different restrictions for different record
 types), create a parameter
 mail_restrict_follower_selection.domain.$your_model.
 
-As an example, you could use [('category_id.name', '=', 'Employees')] to
-allow only contacts with 'Employees' tag to be added as follower - this
-also is the default.
+Some examples:
+
+-  ``[("category_id.name", "=", "Employees")]`` : Only allow contacts
+   with 'Employees' tag
+-  ``[("is_company", "=", False)]`` : Restrict company contacts to be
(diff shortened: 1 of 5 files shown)


AReitenauer:
Functional review ok!
When parameter are configured it works, and saving seams also to work properly.

Only the Info that no restriction is the new default needs to be put up. 
image

And maybe both default system parameter "should" be the same value (empty or 1=1)
image

yostashiro (approved):
Code review and functional test. Looks nice. 👍

AReitenauer (approved):
Functional review ok!
When parameter are configured it works, and saving seams also to work properly.

Only the Info that no restriction is the new default needs to be put up. 
image

And maybe both default system parameter "should" be the same value (empty or 1=1)
image

CRogos:
@pedrobaeza could you merge?

pedrobaeza:
I don't know this module, so some PSC/maintainer of it some review it.

CRogos:
I hope this is the right list? https://oca.github.io/repo-maintainer-conf/teams.html#mail-maintainers
@yajo @etobella ... could you review/merge?

etobella:
@CRogos you can use @OCA/mail-maintainers to simplify :wink:

etobella (approved):
TEchnically looks ok

/ocabot merge minor

PR #146, opened 2026-02-09
[FIX] mail_debrand: debrand notification emails after Odoo layout change

Jord38:
Odoo commit 9f65f1a introduced conditional footer rendering in mail_notification_layout, but mail_debrand only hooked into _render_template (which notification emails bypass via ir.qweb._render) and _prepare_outgoing_body.

Add _notify_by_email_render_layout override on mail.thread to catch branding at the actual render point for all notification emails.

Also clean up leftover pipe separator and whitespace from the unfollow span after removing the Odoo branding link.

### DIFF
--- a/mail_debrand/__manifest__.py
+++ b/mail_debrand/__manifest__.py
@@ -12,7 +12,7 @@
     ( for powerd by) form all the templates
     removes any 'odoo' that are in tempalte texts > 20characters
     """,
-    "version": "18.0.1.0.1",
+    "version": "18.0.1.0.2",
     "category": "Social Network",
     "website": "https://github.com/OCA/mail",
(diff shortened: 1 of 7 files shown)


Larsq1 (approved):

Larsq1:
LGTM

mail_debrand/__manifest__.py
```diff
@@ -12,7 +12,7 @@
     ( for powerd by) form all the templates
     removes any 'odoo' that are in tempalte texts > 20characters
     """,
-    "version": "18.0.1.0.1",
+    "version": "18.0.1.0.2",
```
sergiocorato, line 15: Why this manual change of the version?

Issue #114, opened 2025-11-17
mail_activity_team configuration may lead to unwanted errors


jgebel:
Configuration of activity teams may lead to unwanted errors when activities are created automatically.
Note: If this is indeed the desired behavior, then please see the section "Additional context"

## Module
mail_activity_team

## Describe the bug
You have a user which is not in your activity team and you do something that would automatically create an activity for that user you get the error: "The assigned user <User not in your activity team> is not member of the team <Name of your activity team>."
E.g. when leaves/leave allocations are created, see example below.

## To Reproduce
**Affected versions**:
18.0

1. Start runboat and login as admin
2. Install hr_holiday module (and of course mail_activity_team)
3. Create 2 new users (one will be refered as "base user", the other one as "manager")
4. Set access rights for Time-off
a) For base user leave time-off empty
b) For manager set "Officer: Manage all requests" role
5. For both new users use the action to create employees
6. Go to the employee app and set the following on the employees:
a) For base user: set manager to the created manager user
7. Go to time off app -> configuration -> time off types
8. Choose a type (e.g. "Paid Time Off") and set "Approval" to "By Employee's Approver" (for Time Off Requests and Allocation Requests)
9. Create two Activity teams, e.g. "Admin" with the admin in it and "HR" with the manager user in it.
10. Try to create (as admin) an allocation request for the base user.
The activity, by configuration, should be created for the manager user but the error is thrown mentioning the configuration of the activity team.
(I tried to keep the example simple, see further explanations below)

**Expected behavior**
You should be able to use Odoo default workflows even when the activity team members may differ for the activity creating and receiving users.
(E.g. you have multiple activity teams for HR where every team manages a different department with its employees and the teams may have overlapping responsibilities.)

**Additional context**
If this is really the desired behavior then it would be nice to have an option on the activity teams to not only set the default team but also to set if a team can be assigned for activities of that type.
So when e.g. a leave allocation would be created the activity type "Allocation Approval" has a option like "No team activity" set then the activity would be assigned to the user (like the Odoo default workflow) instead of a team.

lorenzomorandini:
I can confirm we faced the same problem this week

PR #149, opened 2026-02-12
[18.0][IMP] mail_activity_team: Improve default team ID retrieval and update field defaults

CRogos:
Solves: https://github.com/OCA/mail/issues/114

@StefanRijnhart could you have a look in this change?

What do you think about replacing team_user_id by user_id (removing team_user_id)? 
I don't understand why this is needed.

@jgebel @lorenzomorandini could you review/test?

### DIFF
--- a/mail_activity_team/models/mail_activity.py
+++ b/mail_activity_team/models/mail_activity.py
@@ -8,34 +8,26 @@ from odoo.exceptions import ValidationError
 class MailActivity(models.Model):
     _inherit = "mail.activity"
 
-    @api.model
-    def _get_default_team_id(self, user_id=None, model_id=None):
-        if not user_id:
-            user_id = self.env.uid
-        if not model_id:
-            res_model = self.env.context.get("default_res_model")
-            model = (
-                self.sudo().env["ir.model"].search([("model", "=", res_model)], limit=1)
+    def _get_default_team_id(self):
+        self.ensure_one()
+        domain = []
+        domain.append(("member_ids", "=", self.user_id.id or self.env.user.id))
+        if self.res_model_id:
+            domain += [
+                "|",
+                ("res_model_ids", "=", False),
+                ("res_model_ids", "=", self.res_model_id.id),
+            ]
+        if not domain:
+            return self.env["mail.activity.team"]
+        teams = self.env["mail.activity.team"].search(domain)
+        if self.res_model_id:
+            # Prefer teams with a matching model
+            teams = (
+                teams.filtered(lambda mat: self.res_model_id in mat.res_model_ids)
+                or teams
             )
-            model_id = model.id if model else None
-        domain = [("member_ids", "in", [user_id])]
-        if model_id:
-            domain.extend(
-                [
-                    "|",
-                    ("res_model_ids", "=", False),
-                    ("res_model_ids", "in", [model_id]),
-                ]
-            )
-        results = self.env["mail.activity.team"].search(domain)
-        if model_id:
-            result_with_model = results.filtered(
-                lambda team: model_id in team.res_model_ids.ids
-            )
-            if result_with_model:
-                return result_with_model[0]
-
-        return results[0] if results else self.env["mail.activity.team"]
+        return teams[:1]
 
     user_id = fields.Many2one(string="User", required=False, default=False)
     team_user_id = fields.Many2one(
@@ -45,7 +37,26 @@ class MailActivity(models.Model):
     team_id = fields.Many2one(
         comodel_name="mail.activity.team",
         index=True,
+        compute="_compute_team_id",
+        store=True,
+        readonly=False,
     )
+
+    @api.depends("res_model_id", "user_id")
+    def _compute_team_id(self):
+        """Assign team if no team yet or team is incompatible"""
+        for activity in self:
+            if (
+                not activity.team_id
+                or activity.user_id
+                and activity.user_id not in activity.team_id.member_ids
+                or (
+                    activity.res_model_id
+                    and activity.team_id.res_model_ids
+                    and activity.res_model_id not in activity.team_id.res_model_ids
+                )
+            ):
+                activity.team_id = activity._get_default_team_id()
 
     @api.model_create_multi
     def create(self, vals_list):
@@ -63,29 +74,7 @@ class MailActivity(models.Model):
                 # so if we don't have a user_team_id we don't want user_id too
                 if "user_id" in vals and not vals.get("team_user_id"):
                     del vals["user_id"]
-            elif (
-                "team_id" not in vals
-                and "user_id" in vals
-                and "team_user_id" not in vals
-            ):
-                # legacy behavior, if we have user_id but no team_id, set the team_id
-                team_id = self._get_default_team_id(
-                    vals["user_id"], vals.get("res_model_id")
-                )
-                if team_id:
-                    vals["team_id"] = team_id.id
-                    vals["team_user_id"] = vals["user_id"]
         return super().create(vals_list)
-
-    @api.onchange("user_id")
-    def _onchange_user_id(self):
-        if not self.user_id or (
-            self.team_id and self.user_id in self.team_id.member_ids
-        ):
-            return
-        self.team_id = self._get_default_team_id(
-            self.user_id.id, self.sudo().res_model_id.id
-        )
 
     @api.onchange("team_id")
     def _onchange_team_id(self):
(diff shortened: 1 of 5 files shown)


lorenzomorandini:
Hi @CRogos , tested it and it does seem to fix yes (I tested it with the hr_holiday workflow).

lorenzomorandini:
This probably means one thing: server side code that create activities will not have a team_id set by default, even if a matching team exists for that model and user. Is that wanted?

CRogos:
I see what you mean. The current behavior is, that the default Team of the user triggering the activity is used, not the default team of the user assigned to the activity. 
Hard to say what is the best/intended behavior. There is nothing mentioned in the description. All the other modules which might create activities are not aware of team_id, so leaving it empty would be the easiest solution.

On the other hand, we could also try to find the most suitable team for the assigned user, based on model-type and user, and only leave it empty if there is no match, and there is no explicit team set on the activity type.

When a special Activity Type is used, a Default Team can be set, and therefore the team_id would be set. (this could also lead to contraint errors, if the intended user_id is not member of the team.)
image


Personally I also see no need for the constraint, that the user_id must be member of the used team_id. The activity might be in the hands of a team, but there might be reasons that one activity is assigned to someone outside the team, without removing the team. (e.g. vacation replacement or a special customer should not be contacted by the Sales Team, but by the CEO itself [who is not member of the Sales Team])
Nevertheless removing the constraint, will only fix the symptom, but not the problem, that the wrong team is assigend.

I tend to the behavior, try to assign the team based of the assigned user/model and allow setting a team where the assigned user is not part of the team (and removing team_user_id).

lorenzomorandini:
I agree. I don't think it really makes sense to set the team based on the user creating the activity, it should only be based on the assigned user. I also agree on removing the constraint, it can eventually be re-added by a custom module.

StefanRijnhart (commented):
Thanks! Slightly worried about the loss of functionality here, even though I have my own workaround for this which very much amounts to the current change. Would it make sense to update the team if a user is set that is not a member of the current team? That might be possible if we rewrite `team_id` to be a stored, readwrite computed field.

A small number of tests might help to determine the impact of the change. Would you be able to work this out a little further?

CRogos:
@StefanRijnhart @lorenzomorandini I'll give it a try and keep you posted.

CRogos:
@StefanRijnhart @lorenzomorandini Regarding the team_user_id I need to take a deeper look also in the tests. For now you could test the bugfix.

mail_activity_team/models/mail_activity.py
```diff
@@ -8,15 +8,20 @@
...
+            model = (
+                self.sudo().env["ir.model"].search([("model", "=", res_model)], limit=1)
+            )
+            model_id = model.id if model else None
         domain = [("member_ids", "in", [user_id])]
-        if res_model:
+        if model_id:
             domain.extend(
-                ["|", ("res_model_ids", "=", False), ("res_model_ids", "in", model.ids)]
+                ["|", ("res_model_ids", "=", False), ("res_model_ids", "in", model_id)]
```
lorenzomorandini: Wouldn't it be safer to wrap model_id in a list?

lorenzomorandini (changes requested):

StefanRijnhart:
> @StefanRijnhart @lorenzomorandini Regarding the team_user_id I need to take a deeper look also in the tests. For now you could test the bugfix.

I would prefer to take a look when it is finished

mail_activity_team/models/mail_activity.py
CRogos: Agree... but it is also wrong in another way, because if there is a team with and without model set, the team with the model wasn't always returned.

CRogos:
Tests fail due to the use of odoo-test-helper are being fixed in https://github.com/OCA/odoo-test-helper/pull/39

CRogos:
@lorenzomorandini could you have a look?

lorenzomorandini (changes requested):
The new fallback branch in create auto-assigns a team whenever user_id is present and team_user_id is absent, but it does this even when callers explicitly pass team_id=False. That changes prior behavior where an explicit false value suppressed default assignment, so integrations creating non-team activities with {"team_id": False, "user_id": ...} will now get a team unexpectedly. This is a functional regression because it overrides caller intent and can alter downstream routing/visibility logic.

StefanRijnhart:
@lorenzomorandini Thanks for observing. @CRogos This effect would be avoided if you went with the computed field option.

CRogos:
@lorenzomorandini did I add the tests correctly?

lorenzomorandini (changes requested):
Nearly there. When create() receives a payload with team_id=False, a valid user_id, and no team_user_id, this new branch still runs and deletes user_id. That silently creates an unassigned activity even though the caller explicitly set an assignee, which is a regression from the previous if vals.get("team_id") behavior that only removed user_id when a real team was selected.


Regression test
```python
def test_create_keeps_user_when_team_explicitly_false(self):
    """Explicit team_id=False must not drop an explicit assignee."""
    activity = (
        self.env["mail.activity"]
        .with_user(self.employee)
        .sudo()
        .create(
            {
                "activity_type_id": self.activity1.id,
                "note": "Regression check: keep explicit assignee.",
                "res_id": self.partner_client.id,
                "res_model_id": self.partner_ir_model.id,
                "user_id": self.employee2.id,
                "team_id": False,
            }
        )
    )

    self.assertFalse(activity.team_id)
    self.assertEqual(activity.user_id, self.employee2)
```

Proposed fix
```python
if vals.get("team_id"):
    # team choosen: user_id must follow team_user_id
    if "user_id" in vals and not vals.get("team_user_id"):
        del vals["user_id"]
elif "team_id" not in vals and "user_id" in vals and "team_user_id" not in vals:
    # legacy: choose the team only if team_id has not been passed
    team_id = self._get_default_team_id(vals["user_id"], vals.get("res_model_id"))
    if team_id:
        vals["team_id"] = team_id.id
```

lorenzomorandini (approved):
LGTM

CRogos:
@lorenzomorandini thanks for your support.

lorenzomorandini:
Could we get this merged?

CRogos:
@StefanRijnhart could you review and merge?

CRogos:
@muaazsiddiq feel free to review ;)

lorenzomorandini:
@StefanRijnhart let's not forget about this please

mail_activity_team/models/mail_activity.py
```diff
@@ -45,8 +61,19 @@ def create(self, vals_list):
...
+            elif (
+                "team_id" not in vals
+                and "user_id" in vals
+                and "team_user_id" not in vals
+            ):
+                # legacy behavior, if we have user_id but no team_id, set the team_id
+                team_id = self._get_default_team_id(
+                    vals["user_id"], vals.get("res_model_id")
+                )
+                if team_id:
```
lorenzomorandini: Found a little bug, we need to keep the team_user_id

```python
if team_id:
    vals['team_id'] = team_id.id
    if vals.get('user_id'):
        # Keep explicit assignee when team is inferred automatically.
        vals['team_user_id'] = vals['user_id']
```

lorenzomorandini (changes requested):

mail_activity_team/models/mail_activity.py
lorenzomorandini: Without this fix holiday requests would get assigned to the HR team but not to the HR users

CRogos: I think you are right

StefanRijnhart (changes requested):
Functionally agreed now, and great work on the tests! Technically, I still think that a compute method is the way to go. I'm proposing that into this branch in https://github.com/c4a8-odoo/module-oca-mail/pull/3

CRogos:
@StefanRijnhart is it OK for you that I squash all 4 commits?

CRogos:
@StefanRijnhart can you final review/merge?

CRogos:
Tests are failing because of this change: https://github.com/odoo/odoo/commit/9ab602dc232a49f219b6b7cec1de8cf82fb19ca1

CRogos:
@StefanRijnhart can we finalize this PR?

StefanRijnhart (approved):
Thanks for taking my suggestions! Let's see what the tests do now during the merge attempt.

/ocabot merge minor

StefanRijnhart:
CI still fails. Thanks for referring to the upstream change. I see a fix is proposed in https://github.com/odoo/odoo/pull/255720. In the mean time, I'm proposing a workaround in https://github.com/OCA/mail/pull/166.

PR #150, opened 2026-02-15
[19.0][MIG] mail_restrict_follower_selection: Migration to 19.0

muaazsiddiq:


mathiasfrancke (approved):
Coder review: Nothing to complain
Tests: Successfully tested

I found a minor problem: When entering an invalid domain such as "["name", "not like", "%non%"]", the Sales app becomes unusable

mail_restrict_follower_selection/tests/test_mail_restrict_follower_selection.py
```diff
@@ -8,24 +8,25 @@
 
 
 class TestMailRestrictFollowerSelection(TransactionCase):
```
CRogos, line 10: > https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-19.0
> Disable tracking by setting tracking_disable on the whole test env (eg: cls.env = cls.env(context=dict(cls.env.context, tracking_disable=True)) ) or evaluate the use of BaseCommon as base test class if you need to setup company, currency, users and groups in a special way).

mail_restrict_follower_selection/models/mail_thread.py
```diff
@@ -0,0 +1,61 @@
...
+        test_condition = config["test_enable"] and not self.env.context.get(
+            "test_restrict_follower"
+        )
+        if test_condition or self.env.context.get("no_restrict_follower"):
+            return result
+        domain = self.env[
+            "mail.followers.edit"
+        ]._mail_restrict_follower_selection_get_domain()
+        eval_domain = safe_eval(
+            str(domain), context={"ref": lambda str_id: _id_get(self.env, str_id)}
```
CRogos: Can't we use this to remove the utils.py file?

```suggestion
            str(domain), context={"ref": lambda str_id: self.env.ref(str_id).id}
```

mail_restrict_follower_selection/models/mail_followers.py
```diff
@@ -0,0 +1,59 @@
...
+        res_ids,
+        partner_ids,
+        subtypes,
+        check_existing=False,
+        existing_policy="skip",
+    ):
+        test_condition = config["test_enable"] and not self.env.context.get(
+            "test_restrict_follower"
+        )
+        if test_condition or self.env.context.get("no_restrict_follower"):
```
CRogos: Can we remove the **test_condition** because I've changed the default value in this commit? https://github.com/OCA/mail/commit/516c7ad1102949b859a3bf856cdcd033803170d8

CRogos: Can't we remove **test_condition** because I've changed the default value in this commit?
https://github.com/OCA/mail/commit/516c7ad1102949b859a3bf856cdcd033803170d8

CRogos (changes requested):

mail_restrict_follower_selection/models/mail_followers.py
muaazsiddiq: @CRogos Then this should also be removed from the `mail_thread.py` right ?

muaazsiddiq: @CRogos Then this also implies to remove the `test_restrict_follower` parameter from the test context as well. It doesn't hurt to keep it but at the same time it is not being used.

muaazsiddiq:
@CRogos

I have restructure the module a bit by introducing wizard directory. It would be great if you can review it.

CRogos (changes requested):
1) Not sure what is best practice here to either squash the refactoring into the migration commit or leave it as a second commit. But you should squash at least the last two commits.

2) Could you also fix pre-commit?

3) It would also be great if you could squash these kind of commits: 
https://github.com/OCA/maintainer-tools/wiki/Merge-commits-in-pull-requests#mergesquash-the-commits-generated-by-bots-or-weblate
Image

4) Testes with domain [("is_company","=",False)] and was working fine.

MohamedOsman7:
@muaazsiddiq Are you still working on this?

PR #151, opened 2026-02-18
19.0 mig mail quoted reply

mathiasfrancke:


jabelchi:
This feature is a included in the core in 19.0.

mail-reply

PR #153, opened 2026-02-23
[19.0][MIG] mail_activity_team: Migration to 19.0

muaazsiddiq:


mail_activity_team/models/mail_activity.py
```diff
@@ -0,0 +1,107 @@
...
+            # SUPERUSER is used to put mail.activity on some objects
+            # like sale.order coming from stock.picking
+            # (for example with exception type activity, with no backorder).
+            # SUPERUSER is inactive and then even if you add it
+            # to member_ids it's not taken account
+            # To not be blocked we must add it to constraint condition.
+            # We must consider also users that could be archived but come from
+            # an automatic scheduled activity
+            if (
+                activity.user_id.id != api.SUPERUSER_ID
```
mathiasfrancke, line 108: (optional) we could import SUPERUSER_ID from api so we can use `SUPERUSER_ID` here instead of `api.SUPERUSER_ID`

mathiasfrancke (commented):
One optional change. Tests pending

mathiasfrancke:
@muaazsiddiq Some checks are failing

mathiasfrancke (commented):
Tested successfully

CRogos (changes requested):
There are several PR on v18. https://github.com/OCA/mail/pulls?q=is%3Apr+is%3Aopen+mail_activity_team
I think at least the outcome of https://github.com/OCA/mail/pull/116 and https://github.com/OCA/mail/pull/149 should be added before the migration.

Also the changes need to be squashed into one commit and there are also a lot of commits which could be squashed: https://github.com/OCA/maintainer-tools/wiki/Merge-commits-in-pull-requests#mergesquash-the-commits-generated-by-bots-or-weblate

mail_activity_team/tests/test_mail_activity_team.py
```diff
@@ -0,0 +1,429 @@
...
+            {
+                "name": "Call for Demo",
+                "delay_count": 6,
+                "delay_unit": "days",
+                "summary": "ACT 2 : I want to show you my ERP !",
+                "res_model": "res.partner",
+            }
+        )
+        # Create Teams and Activities
+        cls.partner_client = cls.env.ref("base.res_partner_1")
```
CRogos: Demo data is not loaded on tests anymore and need to be replaced by test data.

muaazsiddiq:
@CRogos 
- I have squashed the commit. 
- I will add above mentioned PRs once they get merged.

yaazkal:
> * I will add above mentioned PRs once they get merged.

they are merged now

CRogos (changes requested):
> > * I will add above mentioned PRs once they get merged.
> 
> they are merged now

But thy are now part of the Migration commit which is not a good solution and not following the migration guide.

muaazsiddiq:
> > > * I will add above mentioned PRs once they get merged.
> > 
> > 
> > they are merged now
> 
> But thy are now part of the Migration commit which is not a good solution and not following the migration guide.

Ok, what do you suggest. Should I do the fresh migration with a new PR or is there any other solution that you would suggest ?

CRogos:
I've created a new PR: https://github.com/OCA/mail/pull/177

PR #155, opened 2026-02-26
[17.0][ADD]mail_sent_history

GuillermoNforgeflow:
This module introduces a “Sent History” mailbox in Odoo's messaging system, giving users a dedicated view of all messages and notes they have personally sent. Users can also search through their sent messages using the search bar.


AndreuOForgeFlow (approved):
Interesting feature 💯 LGTM 👍

mail_sent_history/controllers/__init__.py
```diff
@@ -0,0 +1,3 @@
+# Part of Odoo. See LICENSE file for full copyright and licensing details.
+
```
GuillemCForgeFlow:
```suggestion
```
I would remove these copyright statements, it doesn't apply for OCA modules

mail_sent_history/controllers/mailbox.py
```diff
@@ -0,0 +1,24 @@
...
+        domain = [("author_id", "=", partner_id), ("message_type", "in", ["comment"])]
+        res = request.env["mail.message"]._message_fetch(
+            domain,
+            search_term=search_term,
+            before=before,
+            after=after,
+            around=around,
+            limit=limit,
+        )
+        return {**res, "messages": res["messages"].message_format()}
```
GuillemCForgeFlow, line 24: this is never used, right? where is it called?

mail_sent_history/models/mail_message.py
```diff
@@ -0,0 +1,19 @@
...
+
+        messages = self.search(
+            [
+                ("author_id", "=", partner.id),
+                ("message_type", "in", ["comment"]),
+            ],
+            order="date asc",
+        )
+
+        return messages.message_format()
```
GuillemCForgeFlow: this is dangerous as we have no limit here in the search, this could cause perfomance issues in systems where we have lots of records

this set up will also cause the lazy load feature (existing in the other menus) to not work correctly

mail_sent_history/static/src/js/mail_core_web_service.esm.js
```diff
@@ -0,0 +1,29 @@
...
+
+import {MailCoreWeb} from "@mail/core/web/mail_core_web_service";
+import {patch} from "@web/core/utils/patch";
+
+patch(MailCoreWeb.prototype, {
+    setup() {
+        super.setup(...arguments);
+
+        this.messagingService.isReady.then(async () => {
+            const messages = await this.rpc(
```
GuillemCForgeFlow: what about using `orm.call` instead, which is the most standard way

mail_sent_history/static/src/xml/discuss_sidebar_mailboxes.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<odoo>
+
+    <t t-inherit="mail.DiscussSidebarMailboxes" t-inherit-mode="extension">
+        <xpath
+            expr="//div[contains(@class, 'd-flex') and contains(@class, 'flex-column')]"
```
GuillemCForgeFlow: this can break anytime the parent template changes, you could use the `DiscussSidebarMailboxes` root component

GuillemCForgeFlow (changes requested):
the main problem is that it should be following the odoo way to load messages, so we should adapt more to follow odoo standards.

mail_sent_history/controllers/mailbox.py
GuillemCForgeFlow: my bad, it is called here 👀  https://github.com/odoo/odoo/blob/17.0/addons/mail/static/src/core/common/thread_service.js#L161

GuillemCForgeFlow (approved):
functional and code review LGTM 👍🏿

mail_sent_history/models/mail_message.py
```diff
@@ -0,0 +1,23 @@
...
+        ]
+
+        res = self._message_fetch(
+            domain,
+            before=before,
+            after=after,
+            limit=limit,
+        )
+        messages_formatted = res["messages"].message_format()
+        messages_sorted = sorted(messages_formatted, key=lambda m: m["date"])
```
AndreuOForgeFlow:
```suggestion
        messages = res["messages"]._filter_existing_records()
        messages_formatted = messages.message_format()
        messages_sorted = sorted(
            messages_formatted,
            key=lambda m: m["date"] or datetime.min,
        )
```

AndreuOForgeFlow:
```python
def _filter_existing_records(self):
        existing = self.browse()
        for message in self:
            if not message.model or not message.res_id:
                existing += message
            elif self.env[message.model].browse(message.res_id).exists():
                existing += message
        return existing
```

We need to filter out messages of records that have been deleted otherwise users will get an error

AndreuOForgeFlow (changes requested):
Some changes to avoid missing record errors

AndreuOForgeFlow (approved):

AndreuOForgeFlow:
Hi @pedrobaeza, could we perhaps merge this one?

PR #162, opened 2026-03-20
[17.0][ADD] mail_notification_sound_volume

AndreuOForgeFlow:
This module allows users to configure the volume of notification sounds in Odoo (chat messages, inbox notifications, etc.).

A volume slider is added to the user's Preferences page, with a test button to preview the sound at the selected volume.

By default, the volume is set to 100%.


GuillemCForgeFlow (commented):
the idea seems really interesting 👀 but what about renaming to `mail_notification_sound_volume` to be more specific?

mail_notification_sound_volume/static/src/out_of_focus_service_patch.esm.js
```diff
@@ -0,0 +1,29 @@
+/** @odoo-module */
+
+import {OutOfFocusService} from "@mail/core/common/out_of_focus_service";
+import {patch} from "@web/core/utils/patch";
+import {url} from "@web/core/utils/urls";
+
+patch(OutOfFocusService.prototype, {
+    async _playSound() {
```
GuillemCForgeFlow, line 8: just a small issue I see when functionally testing.
1. I play the sound at first, there's no problem there
2. I change the volume on the user preferences
3. I click on the play sound button and it doesn't do anything
4. I click again, and it works

as said, just a minor issue, but we could find the way to solve it before merging

mail_notification_sound_volume/views/res_users_views.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record id="view_users_form_simple_modif_notification_volume" model="ir.ui.view">
+        <field name="name">res.users.preferences.form.notification.volume</field>
+        <field name="model">res.users</field>
+        <field name="inherit_id" ref="base.view_users_form_simple_modif" />
+        <field name="arch" type="xml">
+            <field name="email" position="before">
+                <field name="notification_volume" readonly="0" widget="volume_slider" />
```
GuillemCForgeFlow: perhaps the notification volume only makes sense if the user has the notifications sent via Odoo, right?

GuillemCForgeFlow (changes requested):

mail_notification_sound_volume/static/src/out_of_focus_service_patch.esm.js
GuillemCForgeFlow: I can see this issue seems to be browser related, as the issue only happens to me using Brave:
```
[Brave 1.89.132 (Official Build) (64-bit)](https://brave.com/latest/)
Chromium: 147.0.7727.56
```
therefore, we can close this one, no need to change anything for now. we'll see if the issue persists when used for more users

GuillermoNforgeflow (approved):
lgtm 👍 
Code reviewed and tested

AndreuOForgeFlow:
Hi @etobella & @dreispt, could you review and possibly merge this one?

GuillemCForgeFlow (approved):
lgtm code and functional review 👍🏿

AndreuOForgeFlow:
Hi @pedrobaeza, could we possibly merge this one?

PR #164, opened 2026-03-27
[18.0][ADD] mail_activity_restrict: Option to make activities readonly when user is not assigend

CRogos:
This module gives you the option to restrict the write access on certain activity types. e.g. if activities need to be done by certain users/teams. 

When this option is activated on an activity type, and the activity is not assigned to you, the edit options are hidden.
image

When the activity is assigned to you, it is still visible:
image


CRogos:
@NICO-SOLUTIONS what do you think of this?

NICO-SOLUTIONS:
@CRogos 
I like the idea to prevent activities from being edited or marked as done/cancelled by different users.  👍 
I’ll take a closer look at this in more detail later.

NICO-SOLUTIONS:
@CRogos 
So far, this looks good—just a quick suggestion/question:

It might make sense to restrict editing of activities to the assigned user, and possibly also to the person who created the activity. 

In practice, it happens quite often that someone creates an activity for someone else and later realizes that changes are needed for various reasons. In those cases, it would be helpful if the creator could also edit it.

What do you think?

CRogos:
hmm... I understand your use case. 

In our case, the activity is automatically created, when an equipment is returned. Currently the creator is the person who triggers the workflow (returner), but I assume I can adapt the process that the activity creator can be changed to odoo-bot, because the returner should not be able to reset the activity.

CRogos:
Waiting for review.

mail_activity_team_restrict/models/mail_activity.py
```diff
@@ -0,0 +1,37 @@
...
+
+from odoo import models
+from odoo.exceptions import AccessError
+
+
+class MailActivity(models.Model):
+    """Allow team members to modify restricted activities."""
+
+    _inherit = "mail.activity"
+
```
MohamedOsman7, line 11:
```
def write(self, vals):
        # user looses access to the record when changing the user_id,
        # so we need to check access before changing it
        res = True
        allow_sudo_activity = self.env["mail.activity"]
        if "user_id" in vals:
            allow_sudo_activity = self.filtered(
                lambda act: act._check_access("write") is None
            )
        if allow_sudo_activity:
            res = super(MailActivity, allow_sudo_activity.sudo()).write(vals)
        res = super(MailActivity, self - allow_sudo_activity).write(vals) and res
        return res
```

mail_activity_team_restrict/models/mail_activity.py
```diff
@@ -0,0 +1,37 @@
...
+    def _get_can_write_restrict_allowed_activities(self):
+        """Allow the assignee and members of the assigned team."""
+        return (
+            super()._get_can_write_restrict_allowed_activities()
+            | self.sudo().filtered_domain([("team_id.member_ids", "in", self.env.uid)])
+        )
```
MohamedOsman7, line 17: `return self.sudo().filtered_domain([("user_id", "in", (False, self.env.uid))])`

mail_activity_team_restrict/models/mail_activity.py
```diff
@@ -0,0 +1,37 @@
...
+                )
+            )
+        return AccessError(
+            self.env._(
+                "You cannot %(operation)s some activities because only their assigned "
+                "user or members of the assigned team can modify activities of this "
+                "type.",
+                operation=operation,
+            )
+        )
```
MohamedOsman7, line 37:
```
def _check_access(self, operation: str) -> tuple | None:
        """Restrict write-like operations when the activity type requires it."""
        result = super()._check_access(operation)
        if operation not in ("write", "unlink") or not self or self.env.su:
            return result

        restricted = self.filtered("activity_type_id.can_write_restrict")
        forbidden = restricted - restricted._get_can_write_restrict_allowed_activities()
        if not forbidden:
            return result

        if result and not (forbidden - result[0]):
            return result

        forbidden = result[0] | forbidden if result else forbidden
        return forbidden, lambda: forbidden._make_can_write_restrict_error(operation)
```

MohamedOsman7 (commented):
Everything works as expected. One thing I noticed: When I assign an activity to another user and save, I can no longer reassign it back to myself, from that point on only the assignee has write access. So whoever plans an activity for someone else immediately loses the ability to fix, reassign or cancel it, and no administrator can step in either.

mail_activity_team_restrict/models/mail_activity.py
CRogos: At this point this is an expected behavior. But I agree this could be an issue.
We should add access to the admin group and also add a setting that also grants access for creator.

AnizR (approved):
Code LGTM 👍

PR #165, opened 2026-03-28
[18.0][ADD] mail_activity_reassign_on_archive: New module

rejamen:
This module enhances user archiving by preserving activities instead of deleting them.

When a user is archived in standard Odoo, all their assigned activities are deleted. This module provides an option to reassign those activities to a designated user instead.

Features:
- Configure a fallback user to receive reassigned activities
- Track the original assignee on reassigned activities
- Optionally restore activities when a user is unarchived


ChristophAbenthungCibex (changes requested):
@rejamen 
Overall looks good. Functionality tested and it works well.
The second commit has nothing to do with this PR can you please remove it? (about mail_post_defer https://github.com/OCA/mail/pull/165/changes/280d3b48095cbdc02d87649b72772ded32dd54a1)

mail_activity_reassign_on_archive/models/res_users.py
```diff
@@ -0,0 +1,83 @@
...
+            return
+        activities = (
+            self.env["mail.activity"]
+            .sudo()
+            .search(
+                [
+                    ("user_id", "in", users_to_process.ids),
+                ]
+            )
+        )
```
ChristophAbenthungCibex, line 64: Is the sudo search necessary? It is better to avoid it if it is not necessary. Any user that can archive users should have the rights to write activities.

mail_activity_reassign_on_archive/models/res_users.py
```diff
@@ -0,0 +1,83 @@
...
+        if enabled and restore_enabled:
+            activities = (
+                self.env["mail.activity"]
+                .sudo()
+                .search(
+                    [
+                        ("original_user_id", "in", self.ids),
+                    ]
+                )
+            )
```
ChristophAbenthungCibex, line 81: Same here

ChristophAbenthungCibex (changes requested):

PR #168, opened 2026-03-31
[18.0][FIX] mail_activity_team: remaining calls to previous refactored get_default_team_id method

StefanRijnhart:
Regression of https://github.com/OCA/mail/pull/149

Fixes
```
  File "/mnt/data/odoo-addons-dir/mail_activity_team/wizard/mail_activity_schedule.py", line 53, in _onchange_activity_team_user_id
    and self.activity_team_user_id in self.activity_team_id.member_ids
  File "/opt/odoo/odoo/fields.py", line 3113, in __get__
    return super().__get__(records, owner)
  File "/opt/odoo/odoo/fields.py", line 1244, in __get__
    self.recompute(record)
  File "/opt/odoo/odoo/fields.py", line 1471, in recompute
    apply_except_missing(self.compute_value, recs)
  File "/opt/odoo/odoo/fields.py", line 1444, in apply_except_missing
    func(records)
  File "/opt/odoo/odoo/fields.py", line 1493, in compute_value
    records._compute_field_value(self)
  File "/opt/odoo/odoo/models.py", line 5302, in _compute_field_value
    fields.determine(field.compute, self)
  File "/opt/odoo/odoo/fields.py", line 110, in determine
    return needle(*args)
  File "/mnt/data/odoo-addons-dir/mail_activity_team/wizard/mail_activity_schedule.py", line 30, in _compute_activity_team_id
    ]._get_default_team_id(
TypeError: MailActivity._get_default_team_id() takes 1 positional argument but 3 were given
```

### DIFF
--- a/mail_activity_team/wizard/mail_activity_schedule.py
+++ b/mail_activity_team/wizard/mail_activity_schedule.py
@@ -19,17 +19,28 @@ class MailActivitySchedule(models.TransientModel):
         readonly=False,
     )
 
-    @api.depends("activity_type_id")
+    @api.depends("res_model_id", "activity_user_id")
     def _compute_activity_team_id(self):
-        for scheduler in self:
-            if scheduler.activity_type_id.default_team_id:
-                scheduler.activity_team_id = scheduler.activity_type_id.default_team_id
-            elif not scheduler.activity_team_id:
-                scheduler.activity_team_id = self.env[
-                    "mail.activity"
-                ]._get_default_team_id(
-                    scheduler.activity_team_user_id.id, scheduler.sudo().res_model_id.id
+        """Assign team if no team yet or team is incompatible"""
+        for wizard in self:
+            if (
+                not wizard.activity_team_id
+                or wizard.activity_user_id
+                and wizard.activity_user_id not in wizard.activity_team_id.member_ids
+                or (
+                    wizard.res_model_id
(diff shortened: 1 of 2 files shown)


mail_activity_team/wizard/mail_activity_schedule.py
```diff
@@ -19,17 +19,28 @@ class MailActivitySchedule(models.TransientModel):
...
                 )
+            ):
+                # Reuse mail.activity default team logic
+                activity = self.env["mail.activity"].new(
+                    values={
+                        "res_model_id": wizard.sudo().res_model_id.id,
+                        "user_id": wizard.activity_user_id.id,
+                    },
+                )
+                wizard.activity_team_id = activity._get_default_team_id()
```
CRogos, line 43: I am not convinced that this is a better pattern than adding parameters to the method?

Before activity_team_user_id was used instead of activity_user_id. Is this intendet?

StefanRijnhart: It shouldn't matter whether to use activity_user_id or activity_team_user_id. I believe the dedicated team user fields are only introduced to apply the domain on team membership, but fundamentally the fields are related and kept in sync by Odoo.

What you can do is take the additional test coverage of this PR and copy it into your branch so that we can see how it looks on your PR.

PR #169, opened 2026-04-09
[19.0][MIG] mail_activity_restrict: Migration to 19.0

CRogos:


CRogos:
Feature needs review.

mail_activity_restrict/models/mail_activity.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+class MailActivity(models.Model):
+    """Extend mail.activity access checks for restricted activity types."""
+
+    _inherit = "mail.activity"
+
+    def _get_can_write_restrict_allowed_activities(self):
+        """Return restricted activities the current user may modify."""
+        return self.sudo().filtered_domain([("user_id", "=", self.env.uid)])
```
MohamedOsman7: If nobody is assigned, this matches nobody, so nobody can edit, delete, assign or mark it done, not even Admin. In the UI the buttons are hidden for everyone, and in the Activity overview you only get an AccessError. The activity is stuck forever.

A possible fix:

```
return self.sudo().filtered_domain(
    ["|", ("user_id", "=", False), ("user_id", "=", self.env.uid)]
)
```

So if there is no assignee, normal odoo rules will apply. If assignee is set, the restriction applies.

mail_activity_restrict/models/mail_activity.py
```diff
@@ -0,0 +1,49 @@
...
+            )
+        )
+
+    def _check_access(self, operation: str) -> tuple | None:
+        """Restrict write-like operations when the activity type requires it."""
+        result = super()._check_access(operation)
+        if operation not in ("write", "unlink") or not self or self.env.su:
+            return result
+
+        restricted = self.filtered("activity_type_id.can_write_restrict")
```
MohamedOsman7, line 54: This is actually suggested by Claude and i think it is useful:
"""
`restricted = self.filtered("activity_type_id.can_write_restrict")`

This reads mail.activity.type as the current user inside _check_access. If a read fails here it would raise instead of returning the (records, error) tuple. The helper above does it right with self.sudo(). Works today, but it's fragile:


`restricted = self.sudo().filtered("activity_type_id.can_write_restrict")`

"""

MohamedOsman7 (commented):
@CRogos 

I clicked everything through in the UI. I found these things that i think should be fixed:

1. In connection with the module mail_activity_reminder, nobody can create a restricted activity for other user: If you as Admin, open a contact -> Schedule Activity with "Restrict Write to Assigned User" -> assign it to another user -> Save. You get an AccessError and the activity is not created at all. It only works if you assign it to yourself.

2. An activity without assignee is locked forever: Since 19.0 user_id is optional. ([comment 1](https://github.com/OCA/mail/pull/169/changes#r3767157403)) 

3. ([comment 2](https://github.com/OCA/mail/pull/169/changes#r3767201961))

4. **Should reassigning be allowed at all, and by whom?** Because currently when I create a restricted activity and assign it to myself, then save it, the activity is created successfully. However, if I assign it to someone else, I am no longer able to make any changes to the activity because the buttons are no longer visible. (Yet, with the "mail_activity_reminder" module, you aren't even allowed to change the user, you get an access error.)

mail_activity_restrict/models/mail_activity.py
CRogos: Yes I agree. If nobody is assigned, no restriction should be made. But check this also in the mail_activity_team_restrict. When there is no user_id but a team_id set, in this case only the team should be able to edit.

CRogos:
4. The idea behind the restricted activities is a kind of approval. But approval should only be granted by the assigned users. Therefore changing the user by the creator is not intended. 
When these activity types are defined by a custom module, and the activity is created from code, these activity could also be removed by some code. (like it is done in the timeoff approval process)
We could allow assigning only on create or reassigned by the assigned user/team. Or we could also prevent user from manually creating restricted activities and leave these kind of activity exclusive to activities created by code (like the timeoff approval activity)
image

MohamedOsman7 (approved):
Tested, LGTM

CRogos:
@AnizR could you also have a look into same module v19?

PR #176, opened 2026-04-15
[17.0][ADD] mail_template_domain: New module

cristina-hidalgo-tecnativa:
This module adds a domain filter field on mail templates so only templates matching the active record appear in the email composer dropdown.

@Tecnativa TT61914
@pedrobaeza @victoralmau


mail_template_domain/models/mail_compose_message.py
```diff
@@ -0,0 +1,17 @@
+from odoo import api, fields, models
+
+
+class MailComposeMessage(models.TransientModel):
+    _inherit = "mail.compose.message"
+
+    mail_template_res_id = fields.Integer(
```
pedrobaeza: The filter should be done for all the records, not only the first one. If one of them doesn't comply with the domain, then the template is not shown.

mail_template_domain/models/mail_template.py
```diff
@@ -0,0 +1,33 @@
...
+    _inherit = "mail.template"
+
+    filter_domain = fields.Char(
+        string="Apply on",
+        help="If set, this template will only appear in the email composer "
+        "when the active record matches this domain.",
+    )
+
+    @api.model
+    def _name_search(
```
pedrobaeza: Why overriding this one instead of `search` / `_search`. This one is partial. You should also put a comment as first line explaining why you override the method.

pedrobaeza:
I think you can use `_evaluate_res_ids` for getting the involved records.

mail_template_domain/models/mail_compose_message.py
cristina-hidalgo-tecnativa: Fixed now

mail_template_domain/models/mail_template.py
cristina-hidalgo-tecnativa: `_name_search` is the entry point for Many2one widget lookups. The filter needs Python-side evaluation via `filtered_domain(`) against the actual records, so it can't be pushed down to SQL. Overriding `search/_search` would apply it to all search paths (list views, programmatic calls), which is not the intended behaviour — fail-open outside the composer context is by design.

victoralmau (approved):
Functional review OK

victoralmau:
Ping @pedrobaeza

victoralmau:
Changed to a simpler approach, in my opinion

mail_template_domain/wizards/mail_compose_message.py
```diff
@@ -0,0 +1,38 @@
...
+from odoo.tools.safe_eval import safe_eval
+
+
+class MailComposeMessage(models.TransientModel):
+    _inherit = "mail.compose.message"
+
+    domain_template_id = fields.Binary(compute="_compute_domain_template_id")
+    template_id = fields.Many2one(
+        domain="domain_template_id",
+    )
```
pedrobaeza: Always comment field overrides for not getting confused if it's a new field or what:

```suggestion
    # add domain to the existing field
    template_id = fields.Many2one(domain="domain_template_id")
```

mail_template_domain/wizards/mail_compose_message.py
```diff
@@ -0,0 +1,38 @@
...
+        for item in self:
+            if not item.res_ids:
+                res_ids = []
+            elif isinstance(item.res_ids, list):
+                res_ids = item.res_ids
+            elif isinstance(item.res_ids, str):
+                res_ids = ast.literal_eval(item.res_ids)
+            records = self.env[item.model].browse(res_ids)
+            if not records.exists():
+                item.domain_template_id = []
```
pedrobaeza, line 26:
```suggestion
                item.domain_template_id = []
                continue                
```

pedrobaeza (commented):
Apart from the comments inline, I see the UX is not the best one, as the group is put on top of the settings, where the rest of settings are not related, but not only that, the terminology and the place is not linked with the other field that sets this part. I propose the following:

Image

- Put it after the model selection.
- Add a check for hiding the domain widget (extra boolean field).

victoralmau:
Change done.

mail_template_domain/models/mail_template.py
```diff
@@ -0,0 +1,12 @@
+from odoo import fields, models
+
+
+class MailTemplate(models.Model):
+    _inherit = "mail.template"
+
+    model_filter = fields.Boolean(string="Restrict to")
+    filter_domain = fields.Char(
+        string="Apply on",
```
pedrobaeza:
```suggestion
        string="Restrict domain",
```

mail_template_domain/views/mail_template_views.xml
```diff
@@ -0,0 +1,19 @@
...
+        <field name="model">mail.template</field>
+        <field name="inherit_id" ref="mail.email_template_form" />
+        <field name="arch" type="xml">
+            <field name="model_id" position="after">
+                <field name="model_filter" invisible="not model_id" />
+                <field
+                    name="filter_domain"
+                    widget="domain"
+                    options="{'model': 'model', 'in_dialog': True}"
+                    invisible="not model_filter"
```
pedrobaeza:
```suggestion
                    invisible="not model_filter"
                    nolabel="1"
                    colspan="2"
```

mail_template_domain/wizards/mail_compose_message.py
pedrobaeza: Remove the trailing comma for having it in one line.

mail_template_domain/models/mail_template.py
```diff
@@ -0,0 +1,12 @@
+from odoo import fields, models
+
+
+class MailTemplate(models.Model):
+    _inherit = "mail.template"
+
+    model_filter = fields.Boolean(string="Restrict to")
```
pedrobaeza: For having both fields related by name and proper export labels:

```suggestion
    filter_model = fields.Boolean(string="Restrict?")
```

mail_template_domain/models/mail_template.py
```diff
@@ -0,0 +1,12 @@
...
+    filter_domain = fields.Char(
+        string="Apply on",
+        help="If set, this template will only appear in the email composer "
+        "when the active record matches this domain.",
+    )
```
pedrobaeza, line 15: Make it computed writable for resetting it when `filter_model` is False.

mail_template_domain/views/mail_template_views.xml
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record id="email_template_form_filter_domain" model="ir.ui.view">
+        <field name="name">mail.template.form.filter.domain</field>
+        <field name="model">mail.template</field>
+        <field name="inherit_id" ref="mail.email_template_form" />
+        <field name="arch" type="xml">
+            <field name="model_id" position="after">
+                <field name="model_filter" invisible="not model_id" />
```
pedrobaeza:
```suggestion
                <label for="filter_model" string="Restrict to" invisible="not model_id" />
                <field name="filter_model" nolabel="1" invisible="not model_id" />
```

pedrobaeza (commented):
Almost there, thanks.

pedrobaeza (approved):
/ocabot merge nobump

PR #93, opened 2025-10-15
[18.0][FIX] base_search_mail_content: Fix searching on big messages a…

u0f:
FWP of https://github.com/OCA/social/pull/1734 and fix negative search.

- With these improvements, we fixed the problem of [not being able to search in large texts](https://github.com/OCA/social/pull/1734), giving false negatives. 
- In addition, using the [“any” operator](https://youtu.be/Fmr4JBzlDyY?t=523), the error that [currently exists in v16](https://github.com/OCA/social/pull/1734#issuecomment-3360691728) limited by the ORM is now fixed.      
- New test to support the negative search.   

https://www.loom.com/share/5e0aa4a1ad084ed585b844a5a3e588e7

@moduon MT-10818

PR #88, opened 2025-10-01
[19.0][MIG] mail_message_search: Migration to 19.0

bizzappdev:
As in Odoo, the record_name field is not a stored field anymore. We have added stored true for the record_name field. TBD: Should we remove the search functionality for record_name? Or have it stored=True, in this case, we should already include the migration script and the init hook, in case we are considering record_name as a searchable field.

PR #178, opened 2026-04-15
[MIG] base_search_mail_content: Migrate to 19.0

yajo:
Removed dependency from base_search_fuzzy, which has been unnecessary
since Odoo 16.

Added needed code to keep old behavior.

@moduon MT-14431


yajo:
I wanted to remove the `base_search_fuzzy` because [it should not be needed](https://github.com/OCA/server-tools/tree/15.0/base_search_fuzzy#known-issues--roadmap:~:text=This%20module%20will%20not%20be%20necessary%20from%20version%2016).

While not needed for adding trigram indexes, it still had the nice feature of adding the `%` operator. Still, [that operator did not provide a nice UX](https://github.com/OCA/mail/pull/93#discussion_r2434962629).

Trigram indexes with % operator have [their own problems too](https://github.com/OCA/mail/pull/93), and [doesn't have good multilingual support](https://github.com/OCA/mail/tree/18.0/mail_message_search#use-cases--context:~:text=ineffective%20for%20languages%20without%20space%2Dseparated%20words).

Therefore [I merged `mail_message_search`](https://github.com/OCA/mail/pull/88#issuecomment-4259134205) as a possibly better alternative. At least, it is indeed much simpler, and works for most use cases.

I leave here the code just in case someone wants to take over this module from here. That someone could be my future self, if `mail_message_search` doesn't prove to work for us at the end of the day.

PR #182, opened 2026-04-16
[17.0][ADD] mail_activity_plan_domain: New Module

cristina-hidalgo-tecnativa:
This modules adds a `filter_domain` field to activity plans and their templates. Plans only appear in the scheduling wizard when the active record matches the plan domain. During execution, activities whose template domain does not match the record are silently skipped, allowing a single plan to handle heterogeneous records.

@Tecnativa TT61908
@pedrobaeza @victoralmau


victoralmau (changes requested):
IMO, the best approach is to use the same method that `mail` already has in version 18.0: https://github.com/odoo/odoo/blob/754898c58fded0e3c3c30702f15258f1dadcfa6c/addons/mail/models/mail_activity_plan.py#L30 and add to ROADAMP that it is only necessary up to 17.0; in addition, it would also be necessary to add to OpenUpgrade that this new module is no longer required.

mail_activity_plan_domain/models/mail_activity_schedule.py
```diff
@@ -0,0 +1,116 @@
...
+from odoo.tools.safe_eval import safe_eval
+
+
+class MailActivitySchedule(models.TransientModel):
+    _inherit = "mail.activity.schedule"
+
+    @api.depends("company_id", "res_model_id", "res_ids")
+    def _compute_plan_available_ids(self):
+        """Add res_ids as dependency so that available plans recompute when
+        the selected records change (needed for plan domain evaluation)."""
```
pedrobaeza: On overrides (not original methods), for not replacing docstring:

```suggestion
        # Add res_ids as dependency so that available plans recompute when
        # the selected records change (needed for plan domain evaluation)
```

and the same for the rest.

mail_activity_plan_domain/wizards/mail_activity_schedule.py
pedrobaeza, line 1: This should go into `wizards`, being a transient model.

mail_activity_plan_domain/wizards/mail_activity_schedule.py
```diff
@@ -0,0 +1,116 @@
...
+    _inherit = "mail.activity.schedule"
+
+    @api.depends("company_id", "res_model_id", "res_ids")
+    def _compute_plan_available_ids(self):
+        """Add res_ids as dependency so that available plans recompute when
+        the selected records change (needed for plan domain evaluation)."""
+        return super()._compute_plan_available_ids()
+
+    @api.depends_context("plan_mode")
+    @api.depends("plan_available_ids")
```
pedrobaeza: No need to re-add the depends if they don't change.

```suggestion
```

mail_activity_plan_domain/models/mail_activity_schedule.py
```diff
@@ -0,0 +1,116 @@
+from odoo import _, api, models
+from odoo.tools.safe_eval import safe_eval
+
+
+class MailActivitySchedule(models.TransientModel):
+    _inherit = "mail.activity.schedule"
+
+    @api.depends("company_id", "res_model_id", "res_ids")
```
pedrobaeza: Only add the new dependency. They are joined transversally.

```suggestion
    @api.depends("res_ids")
```

mail_activity_plan_domain/models/mail_activity_schedule.py
```diff
@@ -0,0 +1,116 @@
...
+
+        The core implementation resets plan_id to False unless plan_mode is
+        active. This override keeps the current value when it is still among
+        the available plans, avoiding unwanted resets triggered by the ORM
+        recompute cascade (e.g. when res_ids changes or during create()).
+        """
+        to_recompute = self.filtered(
+            lambda s: not s.plan_id or s.plan_id not in s.plan_available_ids
+        )
+        return super(MailActivitySchedule, to_recompute)._compute_plan_id()
```
pedrobaeza: If you want to make shorter code

```suggestion
        self = self.filtered(
            lambda s: not s.plan_id or s.plan_id not in s.plan_available_ids
        )
        return super()._compute_plan_id()
```

It's not a problem to replace that variable, as it's a local one.

mail_activity_plan_domain/tests/test_mail_activity_plan_domain.py
```diff
@@ -0,0 +1,262 @@
...
+    - Plan domain: plans only appear for matching records.
+    - Template domain: activities are only scheduled for matching records.
+    - Error preview: ignores template domain (shows all potential errors).
+    - Multi-record serialized execution.
+    """
+
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+
```
pedrobaeza: Don't put empty lines inside methods.

pedrobaeza:
@victoralmau I don't get your comment. This one is for having a domain inside the model, not a model restriction.

mail_activity_plan_domain/models/mail_activity_schedule.py
cristina-hidalgo-tecnativa: Modified

mail_activity_plan_domain/wizards/mail_activity_schedule.py
cristina-hidalgo-tecnativa: Modified

mail_activity_plan_domain/wizards/mail_activity_schedule.py
cristina-hidalgo-tecnativa: Removed

mail_activity_plan_domain/models/mail_activity_schedule.py
cristina-hidalgo-tecnativa: Modified

mail_activity_plan_domain/models/mail_activity_schedule.py
cristina-hidalgo-tecnativa: Modified

mail_activity_plan_domain/tests/test_mail_activity_plan_domain.py
cristina-hidalgo-tecnativa: Fixed now

mail_activity_plan_domain/tests/test_mail_activity_plan_domain.py
```diff
@@ -0,0 +1,254 @@
...
+        Uses the same Form-based approach as the core tests to ensure computed
+        fields (plan_available_ids, plan_id) are correctly resolved via
+        onchange semantics before the record is saved.
+        """
+        form = self._instantiate_activity_schedule_wizard(records)
+        if plan:
+            form.plan_id = plan
+        return form.save()
+
+    @users("admin")
```
victoralmau:
```suggestion
```

mail_activity_plan_domain/tests/test_mail_activity_plan_domain.py
```diff
@@ -0,0 +1,254 @@
...
+        self.assertEqual(self.plan_all.domain, "[]")
+        self.assertEqual(self.plan_companies.domain, "[('is_company', '=', True)]")
+        for template in self.plan_mixed.template_ids:
+            self.assertIsNotNone(template.domain)
+
+    @users("admin")
+    def test_plan_available_ids_without_domain(self):
+        """Plan with no domain is available for any record."""
+        wizard = self._make_wizard(self.partner_individual)
+        self.assertIn(self.plan_all, wizard.plan_available_ids)
```
victoralmau, line 120:
```suggestion
        self.assertIn(self.plan_all, wizard.plan_available_ids)
        self.assertNotIn(self.plan_companies, wizard.plan_available_ids)
        self.assertIn(self.plan_mixed, wizard.plan_available_ids)
```

victoralmau (approved):
Code review OK.

These comments are just suggestions:
- I don't think the `users` decorator is necessary.
- It would be helpful to add `assertIn` and `assertNotIn` to all plans to confirm whether or not they are available in the different tests.

mail_activity_plan_domain/tests/test_mail_activity_plan_domain.py
cristina-hidalgo-tecnativa: Removed

mail_activity_plan_domain/tests/test_mail_activity_plan_domain.py
cristina-hidalgo-tecnativa: Added.

pedrobaeza (approved):
We can merge this version too, although we re-review on 18.

/ocabot merge nobump

victoralmau:
Can you try merging it again?

PR #183, opened 2026-04-16
[19.0][MIG] mail_restrict_follower_selection: Migration to 19.0

CRogos:
supersede: https://github.com/OCA/mail/pull/150

- [x] Fix pre-commit C8113: move `mail_followers_edit.py` from `models/` to `wizard/` directory
- [x] Fix test failures round 1: replace `env.ref()` with `set_param()` in tests
- [x] Fix `safe_eval()` API change: `locals_dict=` → positional `context` dict in Odoo 19.0


MohamedOsman7 (approved):
LGTM

MohamedOsman7:
@trisdoan can you please review this PR?

MohamedOsman7:
@BhaveshHeliconia could you please make a review?

mail_restrict_follower_selection/wizard/mail_followers_edit.py
```diff
@@ -0,0 +1,67 @@
...
+    def get_view(self, view_id=None, view_type="form", **options):
+        result = super().get_view(view_id=view_id, view_type=view_type, **options)
+        arch = etree.fromstring(result["arch"])
+        domain = self._mail_restrict_follower_selection_get_domain()
+        eval_domain = safe_eval(
+            str(domain), {"ref": lambda str_id: _id_get(self.env, str_id)}
+        )
+        for field in arch.xpath('//field[@name="partner_ids"]'):
+            field.attrib["domain"] = str(eval_domain)
+        result["arch"] = etree.tostring(arch)
```
BhaveshHeliconia: Should pass encoding="unicode"

mail_restrict_follower_selection/wizard/mail_followers_edit.py
```diff
@@ -0,0 +1,67 @@
...
+                domain_str = str(
+                    self._mail_restrict_follower_selection_get_domain(
+                        res_model=wizard.res_model
+                    )
+                )
+                allowed_partners = self.env["res.partner"].search(
+                    [("id", "in", wizard.partner_ids.ids)]
+                    + safe_eval(
+                        domain_str,
+                        {"ref": lambda str_id: _id_get(self.env, str_id)},
```
BhaveshHeliconia, line 50: What is the purpose of this?

mail_restrict_follower_selection/tests/test_mail_restrict_follower_selection.py
```diff
@@ -0,0 +1,79 @@
...
+                "res_model": "res.partner",
+                "res_ids": str(self.partner.ids),
+                "operation": "add",
+                "partner_ids": [Command.link(self.partner.id)],
+            }
+        )
+        wizard.edit_followers()
+
+    def test_followers_meet(self):
+        self.partner.write({"category_id": self.category_employees})
```
BhaveshHeliconia: setUp already sets this at create time

mail_restrict_follower_selection/tests/test_mail_restrict_follower_selection.py
```diff
@@ -0,0 +1,79 @@
...
+        )
+
+    def test_fields_view_get(self):
+        result = self.env["mail.followers.edit"].get_view(view_type="form")
+        for field in etree.fromstring(result["arch"]).xpath(
+            '//field[@name="partner_ids"]'
+        ):
+            self.assertTrue(field.get("domain"))
+
+    def send_action(self):
```
BhaveshHeliconia: Should be private method

mail_restrict_follower_selection/__init__.py
```diff
@@ -0,0 +1,7 @@
+# Copyright (C) 2015 Therp BV <http://therp.nl>
+# Copyright (C) 2017 Komit <http://www.komit-consulting.com>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from . import models
+from . import utils
```
BhaveshHeliconia:
```suggestion
```
Remove this import as it is used by wizard/mail_followers_edit.py via its own relative import

mail_restrict_follower_selection/demo/ir_config_parameter_demo.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" ?>
+<odoo>
```
BhaveshHeliconia:
```suggestion
<odoo noupdate="1">
```

BhaveshHeliconia (changes requested):

mail_restrict_follower_selection/wizard/mail_followers_edit.py
CRogos: I had to ask the AI myself:

> That lambda injects a ref() helper into safe_eval so domain expressions can resolve XML IDs safely at runtime.
> 
> In this code, safe_eval(str(domain), {"ref": lambda str_id: _id_get(self.env, str_id)}) means: when the evaluated domain contains something like ref("module.external_id"), call _id_get(self.env, str_id) and return the numeric database ID of that record. The helper itself is defined in utils.py and simply does env.ref(id_str).id, so it deliberately returns only the ID, not the recordset.

So I assume this enables you to use ref in the domain: e.g. [('industry_id','=',ref('base.res_partner_industry_A'))]

BhaveshHeliconia: I've verified on runboat and it gives some error. Could you please check on the runboat with this?

mail_restrict_follower_selection/demo/ir_config_parameter_demo.xml
BhaveshHeliconia: This one is still pending.

mail_restrict_follower_selection/wizard/mail_followers_edit.py
```diff
@@ -0,0 +1,67 @@
...
+    _inherit = "mail.followers.edit"
+
+    @api.model
+    def _mail_restrict_follower_selection_get_domain(self, res_model=None):
+        if not res_model:
+            res_model = self.env.context.get("default_res_model")
+        parameter_name = "mail_restrict_follower_selection.domain"
+        parameter_domain = (
+            self.env["ir.config_parameter"]
+            .sudo()
```
BhaveshHeliconia: self.env["ir.config_parameter"].sudo(): Use variable for this to avoid code duplication

BhaveshHeliconia (changes requested):

mail_restrict_follower_selection/demo/ir_config_parameter_demo.xml
CRogos: This is only demo data and not a real configuration. Not sure if we should flag this as noupdate.... OK I had a look into other demo data, you are right.

mail_restrict_follower_selection/wizard/mail_followers_edit.py
CRogos: I've fixed it and also improved the wizard domain filter.

BhaveshHeliconia: Is there any way to avoid overwriting onClickAddFollowers? If not, then please add a comment explaining why we need overwriting.

CRogos: You are totally right. How about this solution, by intercepting the doAction call? (I have to say that I am not an JS expert, but I like this solution from copilot)

mail_restrict_follower_selection/demo/ir_config_parameter_demo.xml
BhaveshHeliconia: - Can you replace `setup()` with following code snippet?
```js
onClickAddFollowers() {
        const originalDoAction = this.action.doAction.bind(this.action);
        const saved = this.action;
        this.action = Object.assign(Object.create(this.action), {
            // The web client filters the loadViews context down to `lang` and
            // keys ending with `_view_ref`. We smuggle the followed record's
            // model through under a `_view_ref`-suffixed key so the wizard's
            // `get_view` can apply the per-model follower domain.
            doAction: (action, options) =>
                originalDoAction(
                    {
                        ...action,
                        context: {
                            ...action.context,
                            restrict_follower_res_model_view_ref:
                                this.props.thread.model,
                        },
                    },
                    options
                ),
        });
        try {
            return super.onClickAddFollowers();
        } finally {
            this.action = saved;
        }
    },
```

- By this, we can narrow scope from `setup()` to `onClickAddFollowers()` and also make use of arrow function.

CRogos: I've changed it even though I am not sure if this is a better solution. But maybe it is easier to understand because it is better linked to the reason of this change.

BhaveshHeliconia (approved):
Code and functional review LGTM!

CRogos:
@pedrobaeza could you merge?

CRogos:
@pedrobaeza could you retry the merge?

PR #200, opened 2026-05-28
[19.0][FW] mail_activity_team: Optional notify for team members

CRogos:
Port forward of https://github.com/OCA/mail/pull/148

Add the option to send notifications to the team members.
image

### DIFF
--- a/mail_activity_team/models/mail_activity.py
+++ b/mail_activity_team/models/mail_activity.py
@@ -3,6 +3,7 @@
 
 from odoo import SUPERUSER_ID, api, fields, models
 from odoo.exceptions import ValidationError
+from odoo.tools.misc import get_lang
 
 from odoo.addons.mail.tools.discuss import Store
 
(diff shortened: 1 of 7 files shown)


MohamedOsman7 (approved):
LGTM, tested.

CRogos:
Add port forward of https://github.com/OCA/mail/pull/148

CRogos:
@StefanRijnhart could you review/merge?

StefanRijnhart:
Is this a migration fix now or a forward port? There is only one commit.

CRogos:
There were some bugs in the migration and the one PR was missing... I probably squashed everything. 
I agree that two different commits would have been better :(

StefanRijnhart (changes requested):
Please create separate PRs for the fix and the backport

CRogos:
It is not a backport. It is both code which was missed out from the migration. I don't see any benefit to move this into two separate PRs. (I do not have the time, so I won't be able to do that.)

AEstLo:
This is just a forward, right? Can you update the description of the PR?

mail_activity_team/models/mail_activity.py
```diff
@@ -130,3 +188,84 @@ def _onchange_activity_type_id(self):
             if self.user_id not in members and members:
                 self.user_id = members[:1]
         return res
+
+    def _to_store_defaults(self, target):
+        values = super()._to_store_defaults(target)
+        values.append(Store.One("team_id", "name"))
+        return values
```
CRogos: Most of the stuff this is a port forward. But I think there are some overlapping changes with https://github.com/OCA/mail/pull/216. Maybe we could merge the other PR first, than I'll rebase and cleanup this PR to only contain the port of https://github.com/OCA/mail/pull/148/ . 

As you already mentioned in the other PR, the module is currently not working and therefore a base + port is also not working.

AEstLo: I agree

CRogos:
@StefanRijnhart rebased and removed the code already merged with #216 .

StefanRijnhart (approved):
Great, thanks! This is now a very clean port of https://github.com/OCA/mail/pull/148. I slightly updated the original comment to reflect this.

CRogos:
@StefanRijnhart can you also merge?

StefanRijnhart:
@AEstLo Is this one OK for you now?

AEstLo (approved):
LGTM (code review, not tested, but the code is similar to 18.0, and I suppose it is widely tested in that version). Thanks!

StefanRijnhart (commented):
/ocabot merge minor

PR #201, opened 2026-05-29
[18.0][FIX] mail_restrict_follower_selection: use most precise follower selection filter for the current model

lorenzomorandini:
This PR aims to fix the current inability to have a global restrictive `mail_restrict_follower_selection.domain` that can get overridden by another specific and more permissive domain for a single model.

Example:
1. the global domain only allows to add employees as followers
2. for HR recruitment candidate model I want to be able add any contact as follower


lorenzomorandini:
@CRogos would you mind giving a look at this? Thanks!

PR #205, opened 2026-06-04
[18.0][FIX] mail_drop_target: drop existing display defined message

LauraCForgeFlow:
When dropping an email file that Odoo fails to import (ignores any email whose Message-Id already exists), the chatter only showed the generic _"An error occured while uploading."_ notification. This fixes that behavior so the actual message is rendered in the UI. It also improves the message itself: reports which record the already-imported copy is attached to.

**Before**
image

**After**
image

A separate commit reads the disable_notify param with str2bool, so a stored "False" value is no longer treated as truthy.

### DIFF
--- a/mail_drop_target/controllers/discuss.py
+++ b/mail_drop_target/controllers/discuss.py
@@ -42,9 +42,8 @@ class AttachmentController(attachment.AttachmentController):
                 return False
             responseData = {"email_upload": 1}
         except AccessError:
-            responseData = {
-                "error": _("You are not allowed to upload an attachment here.")
-            }
+            message = _("You are not allowed to upload an attachment here.")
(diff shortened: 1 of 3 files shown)


GuillemCForgeFlow (approved):
lgtm 👍🏿

AaronHForgeFlow (approved):
Code LG

JordiMForgeFlow (approved):
code LGTM 👍🏼

JoanSForgeFlow (approved):

BhaveshHeliconia (approved):
LGTM!

LauraCForgeFlow:
Hi @pedrobaeza @etobella, can we merge this? Thanks!

PR #207, opened 2026-06-08
[18.0][MIG] mail_activity_plan_domain: Migration to 18.0

victoralmau:
Migration to 18.0

@Tecnativa TT61908


pedrobaeza (approved):
Tested on runboat

mail_activity_plan_domain/readme/USAGE.md
```diff
@@ -0,0 +1,17 @@
+To use this module:
+
+1. Go to **Discuss > Configuration > Activity Plans**.
```
carlos-lopez-tecnativa: I couldn't find this menu. Could you please update the README?

This menu is located under the `Technical` menu, and depending on the installed apps, it may also appear in other places. However, I don't see it in `Discuss`.

Image

Image

mail_activity_plan_domain/models/mail_activity_plan.py
```diff
@@ -0,0 +1,11 @@
+from odoo import fields, models
+
+
+class MailActivityPlan(models.Model):
+    _inherit = "mail.activity.plan"
+
+    domain = fields.Char(
```
carlos-lopez-tecnativa, line 7: When the model is changed, the domain should be cleared to prevent references to fields that do not exist on the new model.

Image

mail_activity_plan_domain/wizards/mail_activity_schedule.py
```diff
@@ -0,0 +1,108 @@
+from odoo import _, api, models
+from odoo.tools.safe_eval import safe_eval
+
+
+class MailActivitySchedule(models.TransientModel):
+    _inherit = "mail.activity.schedule"
```
carlos-lopez-tecnativa, line 6: I think the domain should also be applied to the summary so that activities that do not satisfy the condition are not displayed.

In my example, the "Email" activity is not applicable according to the configured domain, but it is still displayed in the summary.

Image

Image

mail_activity_plan_domain/wizards/mail_activity_schedule.py
```diff
@@ -0,0 +1,108 @@
...
+        ]
+        return [("id", "in", valid_ids)]
+
+    def _is_plan_domain_matching(self, plan, records, eval_context=None):
+        """Return True if the plan has no domain or if at least one of the
+        given records matches the plan's domain."""
+        if not plan.domain or plan.domain == "[]":
+            return True
+        if eval_context is None:
+            eval_context = {"uid": self.env.uid, "user": self.env.user}
```
carlos-lopez-tecnativa: I see this code repeated several times. Could you extract it into a new function, such as `_get_eval_context`, and reuse it throughout the module? Additionally, this would make it easier to extend or override in another module if necessary.

mail_activity_plan_domain/wizards/mail_activity_schedule.py
```diff
@@ -0,0 +1,108 @@
...
+        for record in applied_on:
+            self.with_context(
+                mail_activity_plan_domain_record_id=record.id
+            ).action_schedule_plan()
+        if len(applied_on) == 1:
+            return {"type": "ir.actions.client", "tag": "soft_reload"}
+        return {
+            "type": "ir.actions.act_window",
+            "res_model": self.res_model,
+            "name": _("Launch Plans"),
```
carlos-lopez-tecnativa:
```suggestion
            "name": self.env._("Launch Plans"),
```

mail_activity_plan_domain/wizards/mail_activity_schedule.py
```diff
@@ -0,0 +1,108 @@
...
+            "view_mode": "tree,form",
```
carlos-lopez-tecnativa:
```suggestion
            "view_mode": "list,form",
```

carlos-lopez-tecnativa (changes requested):

mail_activity_plan_domain/wizards/mail_activity_schedule.py
```diff
@@ -0,0 +1,121 @@
...
+        # active. This override keeps the current value when it is still among
+        # the available plans, avoiding unwanted resets triggered by the ORM
+        # recompute cascade (e.g. when res_ids changes or during create()).
+        self = self.filtered(
+            lambda s: not s.plan_id or s.plan_id not in s.plan_available_ids
+        )
+        return super()._compute_plan_id()
+
+    def _get_eval_context(self, eval_context=None):
+        return eval_context or {"uid": self.env.uid, "user": self.env.user}
```
carlos-lopez-tecnativa: Why did you implement it this way instead of keeping it simple?

```suggestion
    def _get_eval_context(self):
        return {"uid": self.env.uid, "user": self.env.user}
```

mail_activity_plan_domain/wizards/mail_activity_schedule.py
```diff
@@ -0,0 +1,121 @@
...
+        ]
+        return [("id", "in", valid_ids)]
+
+    def _is_plan_domain_matching(self, plan, records, eval_context=None):
+        """Return True if the plan has no domain or if at least one of the
+        given records matches the plan's domain."""
+        if not plan.domain or plan.domain == "[]":
+            return True
+        eval_context = self._get_eval_context(eval_context)
+        return bool(records.filtered_domain(safe_eval(plan.domain, eval_context)))
```
carlos-lopez-tecnativa: If this parameter is not optional, there is no need to call `_get_eval_context()`. In this case, we can rely on the parameter directly. Am I missing anything?

```suggestion
    def _is_plan_domain_matching(self, plan, records, eval_context):
        """Return True if the plan has no domain or if at least one of the
        given records matches the plan's domain."""
        if not plan.domain or plan.domain == "[]":
            return True
        return bool(records.filtered_domain(safe_eval(plan.domain, eval_context)))
```

mail_activity_plan_domain/wizards/mail_activity_schedule.py
```diff
@@ -0,0 +1,121 @@
...
+            lambda tpl: self._is_template_domain_matching(tpl, record, eval_context)
+        )
+
+    def _is_template_domain_matching(self, template, record, eval_context=None):
+        """Return True if the template has no domain or if the given record
+        matches the template's domain."""
+        if not template.domain or template.domain == "[]":
+            return True
+        eval_context = self._get_eval_context(eval_context)
+        return bool(record.filtered_domain(safe_eval(template.domain, eval_context)))
```
carlos-lopez-tecnativa: The same here

```suggestion
    def _is_template_domain_matching(self, template, record, eval_context):
        """Return True if the template has no domain or if the given record
        matches the template's domain."""
        if not template.domain or template.domain == "[]":
            return True
        return bool(record.filtered_domain(safe_eval(template.domain, eval_context)))
```

carlos-lopez-tecnativa (changes requested):
Functionally, everything looks good. I just have a few minor comments about the code. Please let me know if I'm missing anything that is important to understanding the logic.

carlos-lopez-tecnativa (approved):
LGTM

PR #209, opened 2026-06-09
[18.0][MIG] mail_template_domain: Migration to 18.0

victoralmau:
Migration to 18.0

Please @pedrobaeza can you review it?

@Tecnativa TT61914


mail_template_domain/readme/USAGE.md
```diff
@@ -0,0 +1,14 @@
+To use this module:
+
+1. Go to **Settings > Technical > Email > Templates**.
+2. Open or create a mail template.
+3. In the **Apply on** field (Settings tab), set the domain that records must
```
pedrobaeza: This readme must be updated with latest labels.

pedrobaeza (approved):

mail_template_domain/models/mail_template.py
```diff
@@ -0,0 +1,20 @@
...
+    filter_domain = fields.Char(
+        string="Restrict domain",
+        help="If set, this template will only appear in the email composer "
+        "when the active record matches this domain.",
+        compute="_compute_filter_domain",
+        store=True,
+        readonly=False,
+    )
+
+    @api.depends("filter_model")
```
carlos-lopez-tecnativa, line 17: When model_id is changed, the domain should be cleared as well.

Image

carlos-lopez-tecnativa (changes requested):
Tested and working from mail.composer.

If I create a server action to send an email, or if I add the action context to a mail.template, the domain is not applied.

I think supporting server actions may be possible, but I understand that it is outside the scope of this module. Could you add this limitation to the roadmap?

Image

Image

carlos-lopez-tecnativa (approved):
LGTM

PR #212, opened 2026-06-18
[18.0][FIX] mail_tracking: Do not hide exception if not managed

ypapouin:
Raise original exception if not managed by this module.

For my use case, a badly migrated module was missing some arguments and no exception was raised because of this module:
image


pedrobaeza (approved):

tarteo (approved):
Makes sense!

pedrobaeza:
Please fw-port it to 19

PR #214, opened 2026-06-23
[18.0][MIG] mail_visible_email

ntsirintanis:
forwardport of https://github.com/OCA/social/pull/1643


mail_visible_email/readme/DESCRIPTION.md
```diff
@@ -1,25 +1,21 @@
...
 as this will be the relevant part, and the partner names are visible on
 other fields.
-
-For technical reasons this module depends on mail_composer_cc_bcc:
-
-- We need the email_bcc field on mail.mail;
-- The module mail_composer_bcc fundamentally changes the workings of the
-  \_send() method on mail.mail. In order not to have to support both
-  methods depending on whether mail_composer_bcc is installed or not, it
-  is easier to just make sure it is installed.
```
thomaspaulb: Right so yes in 18.0 this isn't true anymore, yet, that also thwarts the reason to not have a regular module and a glue module for the bcc addition.

mail_visible_email/models/mail_mail.py
```diff
@@ -1,54 +1,43 @@
...
-        """Do not override existing emails."""
-        self.ensure_one()
-        message = self.mail_message_id
-        preset_emails = self[fieldname]
-        if preset_emails:
-            emails += tools.email_normalize_all(preset_emails)
-        existing = message[fieldname]
-        if existing:
-            emails += existing.split(",")
-        return ",".join(list(set(emails)))
```
thomaspaulb: These abstractions make sense to me, why lose them upon migration?

mail_visible_email/models/mail_mail.py
```diff
@@ -1,54 +1,43 @@
...
+        smtp_session=None,
+        alias_domain_id=False,
+        mail_server=False,
+        post_send_callback=None,
+    ):
+        """Copy email_to, email_cc, email_bcc to mail.message before sending."""
+        for mail in self.filtered(lambda r: r.state == "outgoing"):
+            vals = {}
+            for fname in ("email_to", "email_cc", "email_bcc"):
+                raw = mail[fname]
```
thomaspaulb: I'm just wondering if here, before `super` is called, these fields already all have their right values.

Because it seems like for example [email_bcc](https://github.com/OCA/mail/blob/18.0/mail_composer_cc_bcc/models/mail_mail.py#L49) is only filled upon calling `_prepare_outgoing_list`, which is part of `super._send()`. 

If it isn't, we can of course ourselves do what `mail_composer_cc_bcc` does to fill the fields:

```
        partners_cc_bcc = self.recipient_cc_ids + self.recipient_bcc_ids
        partner_to_ids = [r.id for r in self.recipient_ids if r not in partners_cc_bcc]
        partner_to = self.env["res.partner"].browse(partner_to_ids)
        email_to = format_emails(partner_to)
        email_cc = format_emails_str(self.recipient_cc_ids)
        email_bcc = [r.email for r in self.recipient_bcc_ids if r.email]
```

Or already sneakily call `prepare_outgoing_list` once to get the values, but then we'd have to mind the context change that the function does

mail_visible_email/models/mail_mail.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+class MailMail(models.Model):
+    _inherit = "mail.mail"
+
+    # email_bcc is provided by mail_composer_cc_bcc
+
+    def _send(
+        self,
+        auto_commit=False,
```
NL66278: Maybe we should take the opportunity of the migration to replace all the keyword arguments with `**kwargs` as none of these arguments is referenced in the overriding method and they are passed unchanged.

NL66278 (approved):
👍 Looks good, one suggestion for an improvement on the original.

thomaspaulb (approved):

tarteo (commented):
I'm not sure how I can test this easily. I created a email template with cc and bcc, when I then send a email using the template it will not show up as in the CC and BCC field.

My mailing template:

Image
The message record: 
Image
The email record:

Image

thomaspaulb:
@tarteo At least for the BCC field, it's added to the `mail.mail` record only upon actual sending, which calls [_prepare_outgoing_list](https://github.com/OCA/mail/blob/1ed37eeccc0d23890813b45d0e13ff1456a6c9e1/mail_composer_cc_bcc/models/mail_mail.py#L28) not yet when the mail is created in Outgoing state. It also triggers on mails sent via composer only, any value on the templates is not taken into account AFAIK.

How did you get the mail to send without an outgoing mail server?

PR #219, opened 2026-06-29
[19.0][FIX] mail_tracking: complete message route test payload

pilarvargas-tecnativa:
The previous commit inadvertently missed the changes to the second mocked message_dict used by the same test. This PR completes the fix by applying the same update there as well.

Add the optional mail headers to the second payload as well to keep the test compatible with inherited mail.thread implementations such as mass_mailing.

@Tecnativa TT63167

@pedrobaeza @victoralmau please review


pedrobaeza (approved):
/ocabot merge patch

PR #204, opened 2026-06-04
[18.0][IMP] mail_tracking: Add email score search functionality

tarteo:
This is useful when sending mass mailings and you want to keep your reputation high.
image

PR #220, opened 2026-07-01
[19.0][IMP] mail_tracking: Add email score search functionality (fw port)

tarteo:
fw port of https://github.com/OCA/mail/pull/204

### DIFF
--- a/mail_tracking/tests/test_mail_tracking.py
+++ b/mail_tracking/tests/test_mail_tracking.py
@@ -53,6 +53,13 @@ class TestMailTracking(TransactionCase, MockSmtplibCase):
         self.recipient = self.env["res.partner"].create(
             {"name": "Test recipient", "email": "recipient@example.com"}
         )
+        self.recipient_1 = self.env["res.partner"].create(
+            {"name": "Test recipient 1", "email": "recipient1@example.com"}
+        )
+        self.recipient_2 = self.env["res.partner"].create(
+            {"name": "Test recipient 2", "email": "recipient2@example.com"}
+        )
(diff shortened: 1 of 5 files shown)


mail_tracking/tests/test_mail_tracking.py
```diff
@@ -811,6 +818,63 @@ def test_partner_email_change(self):
...
+    def test_search_on_email_score(self):
+        recipients_domain = [("id", "in", self.recipients.ids)]
+        mail, tracking = self.mail_send(self.recipient.email)
+        tracking.event_create("open", {})
+        mail, tracking = self.mail_send(self.recipient.email)
+        tracking.event_create("open", {})
+
+        mail, tracking = self.mail_send(self.recipient_1.email)
+        tracking.event_create("reject", {})
+
```
pedrobaeza: As you are putting already a comment line, adding an extra empty line forces a lot of vertical scroll for reading the code, so it's convenient to remove such extra empty lines IMO.

pedrobaeza: It's also a question of following the current code style.

tarteo: I can remove them for you. But I like spacing between blocks comment, for me it's faster to see what the comment is about. We should decide in the OCA what style is preferred, and then enforce it using linting if possible.

pedrobaeza: Thanks for considering my comment. For me, the syntax highlighting already acts for that visual distinction, so no need to add that extra lines, and more lines fits in the same screen size.

pedrobaeza (approved):

PR #228, opened 2026-07-07
[19.0][MIG] mail_notify_employee_leave: Migration to 19.0

GuillermoNforgeflow:
**mail_tracking**'s **test_access_tracking_email** failed on this branch because deferred notifications aren't sent synchronously: the test posts a message and immediately checks the mail.tracking.email, which doesn't exist yet when mail_post_defer defers it. (The test only runs when hr is installed, which is the case here)

Fix: mail_post_defer now skips deferring under test_enable unless a test opts in via the test_mail_post_defer context key. Its own tests set that key to keep testing the deferral. No changes needed in mail_tracking.


mail_post_defer/models/mail_thread.py
```diff
@@ -4,13 +4,17 @@
...
 
 
 class MailThread(models.AbstractModel):
     _inherit = "mail.thread"
 
     def _notify_thread(self, message, msg_vals=False, **kwargs):
         """Defer emails by default."""
+        # During tests, don't defer unless a test explicitly opts in
+        if config["test_enable"] and not self.env.context.get("test_mail_post_defer"):
+            return super()._notify_thread(message, msg_vals=msg_vals, **kwargs)
```
GuillemCForgeFlow, line 17: this solution makes sense 👍🏿 otherwise we are deferring also in tests, which we only want to do when explicitly checking in tests, such as in the actual module's tests

GuillemCForgeFlow (approved):
code review lgtm 👍🏿

PR #231, opened 2026-07-08
[18.0][ADD] mail_composer_cc_bcc_mail_tracking: Show recipients of messages in chatter

divad1196:
Glue module between `mail_composer_cc_bcc` and `mail_tracking`.
It displays the _Cc_ and _Bcc_ information on the chatter's message.

By default, it only shows the "_To_" field from mail_tracking.
We can click on it to collapse it and see all the recipients.

Global settings allow to define the default view and allow/disallow toggle.


henrybackman (approved):

mail_composer_cc_bcc_mail_tracking/models/mail_message.py
```diff
@@ -0,0 +1,64 @@
...
+from odoo import models
+
+from odoo.addons.mail.tools.discuss import Store
+
+
+class MailMessage(models.Model):
+    _inherit = "mail.message"
+
+    def tracking_status(self):
+        """Flag each recipient as Cc/Bcc using the partner-based
```
simahawk: comment

simahawk (approved):

mail_composer_cc_bcc_mail_tracking/models/mail_message.py
divad1196: done

PR #232, opened 2026-07-09
[MIG] mail_composer_cc_bcc: Migration to 18.0

nouh-bou:


mail_composer_cc_bcc/tests/test_mail_cc_bcc.py
```diff
@@ -159,54 +159,6 @@ def test_template_cc_bcc(self):
         expecting = self.partner_cc2 + self.partner_bcc
         self.assertEqual(composer.partner_bcc_ids, expecting)
 
-    def test_template_cc_bcc_with_placeholders(self):
```
henrybackman, line 162: Why was this test just removed? It's not applicable anymore?

mail_composer_cc_bcc/__manifest__.py
```diff
@@ -3,7 +3,7 @@
 {
     "name": "Email CC and BCC",
     "summary": "This module enables sending mail to CC and BCC partners in mail composer form.",  # noqa: E501
-    "version": "18.0.1.0.2",
+    "version": "18.0.1.0.1",
```
henrybackman, line 6: Is this on purpose?

ChristophAbenthungCibex:
This module is already migrated to version 18.0 -> #3 
Why do we need another migration?
Wouldn't it be better to implement the changes into the current version to improve it?

simahawk (changes requested):
what is the purpose of this PR? The module is already migrated.

Issue #142, opened 2026-01-28
[18.0] mail_composer_cc_bcc: Mail header X-Odoo-Bcc leaks BCC e-mail adress to recipient


glasl:
## Module
mail_composer_cc_bcc

## Describe the bug
Sending a message with:

To: `mail_to@example.com`
BCC: `mail_bcc@example.com`

the outgoing e-mails will receive an extra header `X-Odoo-Bcc: mail_bcc@example.com`.
This means, that `mail_bcc` will get leaked to `mail_to` in the source code (not shown in the e-mail client).

## To Reproduce
**Affected versions**: [18.0], maybe earlier versions, too

**Steps to reproduce the behavior:**

To & BCC:
1. Send message in chatter (e.g. crm.lead): `To: mail_to@example.com` and `BCC: mail_bcc@example.com`
2. Inspect source code of both incoming e-mails for `X-Odoo-Bcc: mail_bcc@example.com`
3. Both, the recipient's and BCC's e-mail will contain the header `X-Odoo-Bcc: mail_bcc@example.com`

CC & BCC:
1. Send message in chatter (e.g. crm.lead): `CC: mail_cc@example.com` and `BCC: mail_bcc@example.com`
2. Inspect source code of both incoming e-mails for `X-Odoo-Bcc: mail_bcc@example.com`
3. Both, the recipient's and BCC's e-mail will contain the header `X-Odoo-Bcc: mail_bcc@example.com`

**Expected behavior**
I don't know the history of `X-Odoo-Bcc`, but I would not expect this header in any E-mail.
At least not in the E-mail sent to both, `To:` and/or `CC:`.

**Edge Case: To, CC, BCC**
If the message is sent to three contacts, one in To, CC and BCC, the header `X-Odoo-Bcc` will **not** be added to the message

PR #233, opened 2026-07-13
[18.0][FIX] mail_composer_cc_bcc: prevent Bcc leak and Cc-only emails

divad1196:
## Bcc Leak
Due to X-Odoo-Bcc email header, the Bcc were leaking.  
Report in issue: https://github.com/OCA/mail/issues/142 

X-Odoo-Bcc can still be enabled, but is disabled by default.

## Cc-only emails / Empty To header

Emails with no `To` were sent due to missing `lang` information.  
Odoo groups emails using `lang` -> we need to add it as well.

Since the module ensure that we send 1 email per recipients (To+Cc+Bcc),
We can always drop the Cc-only email.

## smtp_to guard

TLDR: if we had duplicate emails in the recipients, it would have caused:
- Bcc to leak
- All recipients would receive a duplicate email

It's fixed by using a list instead of a set. Guards are added.

Explanation:
`recipients` was a `set()` https://github.com/OCA/mail/pull/233/changes#diff-f9cbf73ff985af002f4ecff926acc085c850771dd464b63fa31d8ecbb9fd4470L53
-> It removed the duplicates

Then we checked and popped from `self.env.context.get("recipients")` 
=> If we had duplicate emails, `recipients` would run off 
=> we don't edit the `smtp_to`
=> Odoo will send a mail for all recipients (To+Cc+Bcc)

### DIFF
--- a/mail_composer_cc_bcc/models/mail_mail.py
+++ b/mail_composer_cc_bcc/models/mail_mail.py
@@ -2,6 +2,8 @@
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
 
+import os
+
 from odoo import fields, models, tools
 
 from odoo.addons.base.models.ir_mail_server import extract_rfc2822_addresses
@@ -25,6 +27,17 @@ class MailMail(models.Model):
 
     email_bcc = fields.Char("Bcc", help="Blind Cc message recipients")
 
+    def _expose_bcc_marker(self):
+        """Whether to also add the informational ``X-Odoo-Bcc`` marker header.
+
+        Disabled by default: the marker survives sending and would expose the
+        bcc recipient on every copy. Enable it through the ``expose_x_odoo_bcc``
+        context key or the ``EXPOSE_X_ODOO_BCC`` environment variable.
+        """
+        if self.env.context.get("expose_x_odoo_bcc"):
+            return True
+        return tools.str2bool(os.environ.get("EXPOSE_X_ODOO_BCC") or "", False)
+
     def _prepare_outgoing_list(
         self, mail_server=False, recipients_follower_status=None
     ):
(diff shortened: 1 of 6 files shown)


mail_composer_cc_bcc/models/mail_mail.py
```diff
@@ -33,44 +46,48 @@ def _prepare_outgoing_list(
...
+        # so Odoo's Cc-only email is always a duplicate here.
+        res = [m for m in res if m["email_to"]]
+
+        # The To, Cc headers must be the same on every email, but no record
+        # holds the whole audience: partner_ids is empty for followers, and the
+        # mail.mail of the other langs are unlinked as they are sent.
         partners_cc_bcc = self.recipient_cc_ids + self.recipient_bcc_ids
-        partner_to_ids = [r.id for r in self.recipient_ids if r not in partners_cc_bcc]
-        partner_to = self.env["res.partner"].browse(partner_to_ids)
+        # all_recipients = self._get_all_recipients()
```
henrybackman:
```suggestion
```

henrybackman (approved):

mail_composer_cc_bcc/models/mail_thread.py
```diff
@@ -123,6 +124,18 @@ def _notify_get_recipients_classify(
             customer_data["recipients"] += ids
         return [customer_data]
 
+    def _notify_thread_by_email(self, message, recipients_data, **kwargs):
+        """Pass the whole audience to `_prepare_outgoing_list`, only known here"""
```
simahawk: this should be a comment

simahawk (approved):

simahawk:
@trisdoan ping

mail_composer_cc_bcc/models/mail_thread.py
divad1196: @simahawk agreed. done.

divad1196:
@trisdoan did you get the opportunity to have a look at these changes ?

trisdoan (approved):

PR #234, opened 2026-07-13
[19.0][MIG] Migrate  fetchmail_from_imap_folder

NL66278:
This was originally the module fetchmail_attach_from_folder in the server-tools repository.

~~WIP: I need to clean the history of unrelated commits from server-tools.~~

Solved with: `git-filter-repo --path fetchmail_attach_from_folder --path-rename fetchmail_attach_from_folder:fetchmail_from_imap_folder`.

See: https://github.com/newren/git-filter-repo


fetchmail_from_imap_folder/models/fetchmail_server.py
```diff
@@ -0,0 +1,93 @@
...
+            vals["state"] = "draft"
+        return super().write(vals)
+
+    @api.onchange("server_type", "is_ssl", "object_id")
+    def onchange_server_type(self):
+        result = super().onchange_server_type()
+        self.state = "draft"
+        return result
+
+    def fetch_mail(self, **kwargs):
```
pitrafa, line 87: On Odoo 19 the cron never reaches the folders.

Odoo 17 and 18 ran the cron through the public `fetch_mail()`, so overriding it
was enough. In 19.0 core `_fetch_mails()` now calls the private `_fetch_mail()`
(odoo/odoo `addons/mail/models/fetchmail.py`) and `fetch_mail()` is only the
"Fetch Now" button. This module still overrides `fetch_mail(**kwargs)` only, so
the scheduled job processes INBOX and silently skips every `fetchmail.server.folder`:
the log shows `Fetched 0 email(s) on imap server ...` and no
`start checking for emails in folder ...` line. Pressing "Fetch Now" works, which
is why it is easy to miss. The tests pass because they call `fetch_mail()`
directly; nothing goes through `_fetch_mails()`.

`folders_only` is also ignored on the cron path for the same reason, so a server
configured to read only its folders gets its INBOX processed with the default
model.

Seen in production on 19.0-20260908 with this PR's head (fe49e30b). Suggested fix:

```python
def _fetch_mail(self, **kwargs):
    result = None
    inbox_servers = self.filtered(lambda s: not s.folders_only)
    if inbox_servers:
        result = super(FetchmailServer, inbox_servers)._fetch_mail(**kwargs)
    self.folder_ids.fetch_mail()
    return result
```

plus a test that runs `_fetch_mails()` with the cron context and asserts the
folder fetch is called. Happy to push a commit on top of the branch if you prefer.

PR #235, opened 2026-07-17
[18.0][ADD] fetchmail_incoming_test: emulate email reception

divad1196:
## Context

We don't always want to setup an incoming mail server and/or configure our mail client for our tests.  

## What the module does

This module allows to test how a mail is handled when received in Odoo.

image

The message is then handled correctly


NL66278 (approved):
👍 LGTM

simahawk (approved):

PR #241, opened 2026-08-09
[19.0][IMP] mail_attach_existing_attachment: foldable attachment sections

cliffkujala:
## What

The composer lists every attachment already on the record as one long column of checkboxes. On a record with any mail history that column pushes the Send button off screen, and the filenames a mail thread accumulates (`image001.png`, `image002.png`, ...) are impossible to tell apart.

This groups the composer's attachments under two foldable headers, using the same `hr`/caret header core already uses for Planned Activities:

- **Available Attachments** is the existing picker. It starts folded, its header counts what is on offer (or what is ticked, once folded back), and unfolded it lays the checkboxes out in as many columns as the composer is wide enough for. The grid is `repeat(auto-fill, minmax(15rem, 1fr))`, so the column count follows the dialog's own width rather than the viewport's.
- **New Attachments** shows the files added to this mail as the same cards the chatter uses (`AttachmentList`), instead of the plain list of links Odoo 19 renders. It only appears once there is at least one file.

The picker also loads its choices with `search_read` rather than `name_search`, whose default limit of 100 silently dropped the oldest attachments of a busy record, and shows each file's size.

Only the comment-mode composer is touched. Mass mail keeps the stock widget, which is consistent with `can_attach_attachment` already being False there.

## Test plan

- [ ] Open the full composer on a record with several attachments: Available Attachments is folded and badged with the count.
- [ ] Unfold it and resize the dialog: the checkboxes reflow between 1, 2 and 3 columns.
- [ ] Tick a couple, fold it back: the badge shows how many are ticked, and sending still attaches them.
- [ ] Attach a file: New Attachments appears with a card, and its remove button still deletes the upload.

### DIFF
--- a/mail_attach_existing_attachment/README.rst
+++ b/mail_attach_existing_attachment/README.rst
@@ -55,6 +55,17 @@ To use this module, you need to:
 
 |image2|
 
+The composer groups attachments under two headers, each folding away
+when its title is clicked:
+
+- **New Attachments** lists the files added to this email as cards, and
+  only appears once there is at least one.
+- **Available Attachments** is the picker for the files already on the
+  record. It starts folded, and its header shows how many files are on
+  offer, or how many are ticked once you fold it back. Unfolded, it lays
+  the checkboxes out in as many columns as the composer is wide enough
+  for.
+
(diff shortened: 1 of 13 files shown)


cliffkujala:
Closing: this went up before the change had been tried in a browser, and it needs another round. Will reopen once verified.

PR #242, opened 2026-08-09
[19.0][IMP] mail_attach_existing_attachment_account: foldable attachment picker

cliffkujala:
## What

Offer the invoice's attachments through the same picker the mail composer gets in #243: folded by default, and laid out in as many columns as the send wizard is wide enough for. An invoice that has been mailed a few times otherwise buries the wizard's buttons under a column of checkboxes.

The account wizard's own "Attach a file" widget is deliberately left alone; only the existing-attachment picker changes.

## Depends on

#243, which adds the `attachment_checkbox_grid` widget in `mail_attach_existing_attachment`. This PR is a one-line widget swap on top of it.

## Test plan

- [x] Open Send & Print on an invoice with several attachments: the picker is folded, and unfolds into a multi-column grid.
- [x] Tick one and send: it is still attached to the outgoing mail.

### DIFF
--- a/mail_attach_existing_attachment_account/README.rst
+++ b/mail_attach_existing_attachment_account/README.rst
@@ -36,6 +36,10 @@ This Module adds the mail_attach_existing_attachment feature to the
 account module, since there is a change in odoo 12.0, which adds a new
 mail wizard form to the account module
 
+The invoice's attachments are offered through the same foldable,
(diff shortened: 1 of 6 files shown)


cliffkujala:
Closing along with #241, which this depends on. Will reopen once verified.

PR #243, opened 2026-08-09
[19.0][IMP] mail_attach_existing_attachment: foldable attachment sections

cliffkujala:
## What

The composer lists every attachment already on the record as one long column of checkboxes. On a record with any mail history that column pushes the Send button off screen, and the filenames a mail thread accumulates (`image001.png`, `image002.png`, ...) are impossible to tell apart.

This groups the composer's attachments under two foldable headers, using the same `hr`/caret header core already uses for Planned Activities:

- **Available Attachments** is the existing picker. It starts folded, its header counts what is on offer (or what is ticked, once folded back), and unfolded it lays the checkboxes out in as many columns as the composer is wide enough for. The grid is `repeat(auto-fill, minmax(min(15rem, 100%), 1fr))`, so the column count follows the dialog's own width rather than the viewport's.
- **New Attachments** shows the files added to this mail as the same cards the chatter uses (`AttachmentList`), instead of the plain list of links Odoo 19 renders. It only appears once there is at least one file.

The picker also loads its choices with `search_read` rather than `name_search`, whose default limit of 100 silently dropped the oldest attachments of a busy record, and shows each file's size.

Only the comment-mode composer is touched. Mass mail keeps the stock widget, which is consistent with `can_attach_attachment` already being False there.

Two notes for reviewers:

- Both widgets set `--fieldWidget-display: block`, as core's `mail_composer_attachment_list` does. Without it the field wrapper is `inline-block`, the header rules have nothing to grow into and the grid cannot get past one column.
- The `min()` in the grid template is interpolated (`#{"min(15rem, 100%)"}`), because libsass otherwise evaluates it as a Sass function and fails on comparing `rem` with `%`.

## Test plan

- [x] Verified in a browser: Available Attachments folded with its count, unfolding into a grid that reflows 3/2/1 as the dialog is resized; New Attachments showing cards and hidden while empty.
- [x] `web.assets_backend` compiles with the new SCSS.

### DIFF
--- a/mail_attach_existing_attachment/README.rst
+++ b/mail_attach_existing_attachment/README.rst
@@ -55,6 +55,17 @@ To use this module, you need to:
 
 |image2|
 
+The composer groups attachments under two headers, each folding away
+when its title is clicked:
+
+- **New Attachments** lists the files added to this email as cards, and
+  only appears once there is at least one.
+- **Available Attachments** is the picker for the files already on the
+  record. It starts folded, and its header shows how many files are on
+  offer, or how many are ticked once you fold it back. Unfolded, it lays
+  the checkboxes out in as many columns as the composer is wide enough
+  for.
+
 .. |image1| image:: https://raw.githubusercontent.com/OCA/mail/19.0/mail_attach_existing_attachment/static/description/attachment.png
 .. |image2| image:: https://raw.githubusercontent.com/OCA/mail/19.0/mail_attach_existing_attachment/static/description/ex_mail_compose_message.png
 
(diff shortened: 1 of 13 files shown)


pedrobaeza:
Please put screenshots. Is this AI assisted? You should indicate it if so. Another thing: this seems more `[IMP]` than `[FIX]`.

cliffkujala:
I agree this is more [IMP] than fix.  Screen shots below.  Is there really much code being created these days that isn't AI assisted? 
Screenshot 2026-08-10 at 06 08 42
Screenshot 2026-08-10 at 06 09 14
Screenshot 2026-08-10 at 06 09 48
Screenshot 2026-08-10 at 06 10 14
Screenshot 2026-08-10 at 06 14 58

cliffkujala:
@pedrobaeza did you have a chance to review the screenshots, or try this modification out?  We are using in our operation and so far my users are loving it.

pedrobaeza (commented):
IMO, too many changes for a small UI improvement. The module is increasing a lot the amount of code, so doing it harder to maintain.

PR #247, opened 2026-08-12
[16.0][ADD] mail_message_history: list the chatter messages in a menu

smorita7749:
This module lists the chatter messages in a dedicated *Discuss > Message History*
menu, where they can be searched, filtered and grouped.

The chatter mixes the messages posted by users with the ones logged by the system
(field tracking, automated notifications), which makes a long history tedious to
follow. On top of that, the tracked field changes are only readable from the
chatter, as `tracking_value_ids` is restricted to `base.group_system`.

The list therefore shows, as plain text, a preview of the message body and a
summary of the tracked field changes, e.g. `Status: RFQ -> Purchase Order`, and
provides filters telling the messages posted by users apart from the ones logged
by the system.

Notes:

* the views are variants of the standard message views (inherited in primary
  mode), so the standard Messages list is left unchanged;
* a button opening the list for a single record is added on the form views of the
  models listed in the `mail_message_view.button_models` system parameter;
* the menu is restricted to a dedicated group, as listing the messages of every
  document is an expensive query.

Targeting 16.0 as this is the series where the module is used, but I am happy to
forward-port it first if maintainers prefer new modules to land on the latest
branch.

This is a new module of about 1000 lines, which is above the 500 lines of the OCA
metrics red lines. Please let me know how you would like to proceed with the
review.

Assisted-by: Claude Opus 5

@qrtl QT7086


mail_message_view/models/mail_message.py
```diff
@@ -0,0 +1,81 @@
...
+
+from odoo import _, api, fields, models
+from odoo.tools.misc import format_date, format_datetime
+
+
+class MailMessage(models.Model):
+    _inherit = "mail.message"
+
+    document_name = fields.Char(string="Document", compute="_compute_document_name")
+    body_preview = fields.Char(compute="_compute_body_preview")
```
AungKoKoLin1997, line 13: Can't we use odoo standard field `preview`?

mail_message_view/models/mail_message.py
```diff
@@ -0,0 +1,81 @@
...
+from collections import defaultdict
+
+from odoo import _, api, fields, models
+from odoo.tools.misc import format_date, format_datetime
+
+
+class MailMessage(models.Model):
+    _inherit = "mail.message"
+
+    document_name = fields.Char(string="Document", compute="_compute_document_name")
```
AungKoKoLin1997, line 12: `document_name` shadows the stored `record_name` with an unstored compute, so the column can't be sorted, searched or grouped, and it costs a display_name read per model per page. The gap is real but it's a core inconsistency — message_post fills record_name while _message_log/_message_log_batch pass 'record_name': False, skipping the backfill in _message_create — so it's better closed by filling the field on create (batched per model) than by computing it on read.

mail_message_view/__manifest__.py
yostashiro, line 1: Please add yourself as a maintainer.

mail_message_view/views/mail_message_views.xml
```diff
@@ -0,0 +1,161 @@
...
+        <field name="mode">primary</field>
+        <field name="priority">100</field>
+        <field name="arch" type="xml">
+            <field name="subject" position="before">
+                <field name="body_preview" />
+                <field name="tracking_summary" />
+            </field>
+            <!-- The subject is empty on most of the chatter messages. -->
+            <field name="subject" position="attributes">
+                <attribute name="optional">hide</attribute>
```
yostashiro, line 20: Shouldn't we flip it to show?

mail_message_view/views/mail_message_views.xml
```diff
@@ -0,0 +1,161 @@
...
+                <field name="message_type" />
+                <field name="subtype_id" optional="hide" />
+                <field name="is_internal" optional="hide" />
+            </field>
+            <field name="model" position="attributes">
+                <attribute name="optional">hide</attribute>
+            </field>
+            <field name="res_id" position="attributes">
+                <attribute name="optional">hide</attribute>
+            </field>
```
yostashiro, line 35: I think this will be useful.
```suggestion
            </field>
            <field name="res_id" position="after">
                <button
                    name="action_open_document"
                    type="object"
                    icon="fa-file-text-o"
                    title="Open Document"
                    attrs="{'invisible': ['|', ('model', '=', False), ('res_id', '=', 0)]}"
                    optional="show"
                />
            </field>
```

mail_message_view/models/mail_message.py
```diff
@@ -0,0 +1,81 @@
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
```
yostashiro, line 1:
```suggestion
# Copyright 2026 Quartile (https://quartile.co)
# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
```

Do the same to other files.

yostashiro (commented):
Partial review.

yostashiro:
@smorita7749 Can we rename the module to mail_message_history?

mail_message_view/models/mail_message.py
smorita7749: Done — dropped `body_preview` for core's `preview`. Its compute is the same call (`_get_message_preview()`) on the same dependency, so the field was a plain duplicate.

mail_message_view/models/mail_message.py
smorita7749: Good call — I went with filling the field on create. `document_name` and its compute are gone, and the list now shows core's `record_name` as *Document*, which also makes it sortable, searchable and groupable.

Two deviations from your suggestion:

- `message_notify` is excluded (`message_type != 'user_notification'`). It goes through `_notify_thread`, and the empty `record_name` is what keeps the subject and the subtitle off its notification email, so filling it there would change outgoing mails.
- Filling on create only covers new messages. Backfilling the existing ones needs a one-off pass resolving `display_name` per document, which I would rather keep out of this PR; the limitation is documented in `readme/ROADMAP.rst`.

mail_message_history/__manifest__.py
smorita7749: Added, and I put Quartile in `author` as well, to match the copyright headers.

mail_message_view/views/mail_message_views.xml
smorita7749: Flipped to `show`. The old comment was note-centric: the menu defaults to user messages, which include emails, and those do have a subject.

mail_message_history/views/mail_message_views.xml
```diff
@@ -0,0 +1,161 @@
...
+                <field name="message_type" />
+                <field name="subtype_id" optional="hide" />
+                <field name="is_internal" optional="hide" />
+            </field>
+            <field name="model" position="attributes">
+                <attribute name="optional">hide</attribute>
+            </field>
+            <field name="res_id" position="attributes">
+                <attribute name="optional">hide</attribute>
+            </field>
```
smorita7749: Added, thanks. `action_open_document` already exists on core `mail.message`, so no new method was needed.

I dropped `optional="show"` though: the list arch parser only reads `optional` on `field` nodes, so on a button it has no effect.

mail_message_history/models/mail_message.py
```diff
@@ -0,0 +1,81 @@
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
```
smorita7749: Added to the manifest and the Python files. I left the XML files out, as we do not carry headers there in our other modules.

smorita7749:
Renamed to `mail_message_history`, including the `mail_message_history.button_models` system parameter. The branch name still says `mail_message_view`, as it cannot be changed without closing the PR.

I kept the view record ids (`mail_message_view_tree` / `_form` / `_search`) as they are: that is the standard `<model>_view_<type>` naming, which core itself uses for `mail.mail_message_view_form`.

The rest of the review is addressed in the individual threads. Force-pushed as a single commit.

mail_message_history/models/mail_message.py
```diff
@@ -0,0 +1,101 @@
...
+            names.update(
+                {(model_name, record.id): record.display_name for record in records}
+            )
+        for values in todo:
+            name = names.get((values["model"], values["res_id"]))
+            if name:
+                values["record_name"] = name
+
+    @api.depends("tracking_value_ids")
+    def _compute_tracking_summary(self):
```
yostashiro, line 68: I tend to think we should remove this and simply not show the tracking records in the message history list view so the scope of this module is more focused. Do we have concrete business needs to keep it?

PR #249, opened 2026-08-17
[18.0][FIX] mail_composer_cc_bcc: email_cc type

trisdoan:
# Context

- Earlier, `mail_composer_cc_bcc` overrides `mail.mail._prepare_outgoing_list()` and was putting a comma-joined string into `email_to`, where native produces a list. 
- Hence, `mail_tracking` re-joins `email_to` with `COMMASPACE`, which mangles a value that is already a string.
- PR #54 fixed it by making `format_emails()` return the list form

In the same commit, in the same function, `email_cc` was changed in the opposite direction — from the list form to the joined-string form:

```python
 email_to = format_emails(partner_to)                 # list   — fixed by #54
-email_cc = format_emails(self.recipient_cc_ids)      # list   — was correct
+email_cc = format_emails_str(self.recipient_cc_ids)  # string — regression
```

So the commit made `email_to` a list and `email_cc` a string, in adjacent lines of the same dict.

# Solution
restoring the list form so `email_cc` matches `email_to` on the line directly

### DIFF
--- a/mail_composer_cc_bcc/models/mail_mail.py
+++ b/mail_composer_cc_bcc/models/mail_mail.py
@@ -10,6 +10,7 @@ from odoo.addons.base.models.ir_mail_server import extract_rfc2822_addresses
 
 
 def format_emails(partners):
+    """List form, for the outgoing-dict slots (e.g. email_cc, email_to)."""
     return [tools.formataddr((p.name or "", p.email)) for p in partners if p.email]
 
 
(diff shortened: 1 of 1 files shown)


trisdoan:
Hello @norlinhenrik , could you take a look please?

norlinhenrik:
```
        email_cc = format_emails(self.recipient_cc_ids)
        email_bcc = [r.email for r in self.recipient_bcc_ids if r.email]
```

Maybe `email_bcc` should be similar to `email_cc`?

Please install `mail_tracking` also, and send an email to multiple CC and multiple BCC. If you confirm that it works, I will approve this PR.

PR #258, opened 2026-08-25
[17.0][ADD] mail_attachment_office_preview: preview Office attachments via LibreOffice conversion

alan196:
Replaces OCA/knowledge#635: moved here and renamed following the same naming criteria as OCA/mail#257 (@pedrobaeza's suggestion), since the module is about mail attachments and is not a derivative of `attachment_preview`.

## Summary

New module that previews office documents (DOCX, XLSX, PPTX, DOC, XLS, PPT, ODT, ODS, ODP, ODG) in the browser — including chatter attachments — by converting them to PDF **server-side** with LibreOffice headless and rendering the result with Odoo's native PDF.js viewer.

Nothing leaves the server: no external viewer service (Microsoft Office Online, Google Docs) is involved, which makes it suitable for sensitive documents.

### Context

- `attachment_preview` was never migrated to 17.0 and its ViewerJS-based approach only covers PDF + ODF anyway.
- OCA/knowledge#603 adds Office-format preview to `attachment_preview` on 18.0 via LibreOffice conversion. This module brings the same conversion-endpoint logic to 17.0, but as a small standalone module: on 17.0 the native FileViewer already renders PDFs, so no ViewerJS or base module is needed.

### How it works

- HTTP endpoint `GET /mail_attachment_office_preview/to_pdf?model=&field=&id=&filename=` converts the binary field content to PDF with `libreoffice --headless --convert-to pdf` (isolated `UserInstallation` profile per conversion, checksum-keyed disk cache under the filestore).
- A JS patch on the mail `Attachment` model marks office extensions as previewable (`isPdf`) and routes `defaultSource` through the conversion endpoint wrapped in the native PDF.js viewer. The *Download* button still returns the original file.
- Endpoint enforces: record read ACL, binary-field allow-list, extension allow-list, 50 MB size cap, 60 s conversion timeout. Returns 503 gracefully when LibreOffice is not installed.
- No new Python dependencies (stdlib `subprocess`/`tempfile` only).

### Test plan

- Unit tests covering the conversion helper and every HTTP branch of the endpoint (LibreOffice is mocked, so tests pass on runners without it).
- Manually validated on Odoo.sh (LibreOffice 7.3.7): xlsx/docx/pptx attachments preview correctly in the chatter.


pedrobaeza:
This module can't be tested, as the LibreOffice headless server is not included. Isn't this another way to preview them?

alan196:
Added `"external_dependencies": {"deb": ["libreoffice"]}` to the manifest, as in `report_py3o`. Runboat installs the declared `deb` dependencies when the build starts, so LibreOffice should now be available and the preview testable there.

alan196:
The red CI here is not caused by this PR: both failures are in `mail_notify_employee_leave`, whose tests only pass while the runner clock falls inside the employee's working schedule. Fixed in #260; once that is merged this PR should be green.

pedrobaeza (approved):
Tested on runboat and working. Please squash both commits into one.

PR #260, opened 2026-08-28
[17.0][FIX] mail_notify_employee_leave: make absence tests time independent

alan196:
`TestNotifyEmployeeLeave` fails depending on the time of day the CI runs.

`_create_validated_leave()` creates a full-day leave (00:00-23:59), but `hr.leave`
narrows it to the working schedule of the employee, so with the standard company
calendar (08:00-17:00 Europe/Brussels) the stored leave is 06:00-15:00 UTC. The
tests then rely on `employee.is_absent`, which is computed as
`date_from <= now <= date_to`, so they only pass while the runner clock happens
to be inside that window.

Reproduced locally: with the clock frozen at 20:00 the two tests fail exactly as
in CI; inside the window they pass. Recent runs match: the 17.0 post-merge run at
07:10 UTC was green, PR runs at 19:20 and 21:02 UTC were red (e.g. #258).

Fix: freeze the clock at a working hour of a working day, so the tests no longer
depend on when they run.

### DIFF
--- a/mail_notify_employee_leave/tests/test_mail_notify_employee_leave.py
+++ b/mail_notify_employee_leave/tests/test_mail_notify_employee_leave.py
@@ -9,13 +9,14 @@ from freezegun import freeze_time
 from odoo import fields
 from odoo.tests.common import TransactionCase
 
-
 # ``is_absent`` is only true while the current time is inside the leave, and
 # ``hr.leave`` narrows a full day leave to the working schedule of the employee
-# (08:00-17:00 in the timezone of the company calendar). Freeze the clock inside
-# that window, otherwise the result depends on the time of the day at which the
-# tests happen to run.
-@freeze_time("2023-05-15 10:00:00")
+# (08:00-17:00 in the timezone of the company calendar). The tests that need an
+# absent employee freeze the clock inside that window, otherwise the result
+# depends on the time of the day at which they happen to run.
+WORKING_HOUR = "2023-05-15 10:00:00"
+
+
 class TestNotifyEmployeeLeave(TransactionCase):
     @classmethod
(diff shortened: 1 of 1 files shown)


mail_notify_employee_leave/tests/test_mail_notify_employee_leave.py
```diff
@@ -4,10 +4,18 @@
...
 from odoo import fields
 from odoo.tests.common import TransactionCase
 
 
+# ``is_absent`` is only true while the current time is inside the leave, and
+# ``hr.leave`` narrows a full day leave to the working schedule of the employee
+# (08:00-17:00 in the timezone of the company calendar). Freeze the clock inside
+# that window, otherwise the result depends on the time of the day at which the
+# tests happen to run.
+@freeze_time("2023-05-15 10:00:00")
```
pedrobaeza: At class level, there are problems with this having side effects. Please do it at method level (including if neeeded `setUpClass`).

alan196: Done, moved to method level in 939d3608.

`@freeze_time` is now applied only on the two tests that need the employee to be absent (`test_notify_when_user_absent` and `test_only_one_notification_per_day`), with the timestamp in a `WORKING_HOUR` constant. `setUpClass` is no longer frozen, so the setup has no side effects.

`_create_validated_leave()` is no longer a `@classmethod`, as it's now called from the test methods (inside the frozen window) instead of the class setup.

Tests pass locally at a time of day that used to break them (23:17 UTC): `0 failed, 0 error(s) of 3 tests`.

pedrobaeza (approved):
/ocabot merge patch

PR #262, opened 2026-09-04
[18.0][FIX] mail_tracking: Retry wizard was empty when reopened

CarlosRoca13:
Reopening the resend wizard on a failed message showed no recipients at all, leaving no way to retry the delivery.

The wizard deduplicated its own lines with a search over the whole mail.resend.partner table instead of scoping it to the current wizard. Those transient records survive between openings until the vacuum removes them, so the second time the wizard was opened the search matched the records created by the first one and every tracking was skipped.

cc @Tecnativa

ping @pedrobaeza @christian-ramos-tecnativa

### DIFF
--- a/mail_tracking/models/mail_message.py
+++ b/mail_tracking/models/mail_message.py
@@ -305,6 +305,25 @@ class MailMessage(models.Model):
         ]
         return res
 
+    def _message_notifications_to_store(self, store: Store):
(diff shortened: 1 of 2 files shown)


pedrobaeza (approved):
Code review. I wonder if we can reuse existing `mail.notification`, but this will be maybe more hackish, so let's continue with the creation of a new one.

PR #263, opened 2026-09-04
[18.0][IMP] mail_tracking: Reactive tracking status in the chatter

CarlosRoca13:
The tracking status shown in the chatter only refreshed on a page reload.

A mail.tracking.email state change happens outside of any user session and doesn't touch mail.notification, so core's _notify_message_notification_update() is never triggered for it and nothing is pushed to the bus. Notify the involved users on every state write instead. The store has to be built once per recipient because is_failed_message is computed for the reader, and the refresh is aggregated in cr.precommit so that a message with several recipients emits a single notification instead of one per email sent.

Furthermore,  MessageTracking captured props.partner_trackings in setup(), which isn't re-run when the props change. As tracking_status() returns a brand new list on every call, the component kept rendering the trackings the message had when it was created. Read them from the props on each render.

cc @Tecnativa 

ping @christian-ramos-tecnativa @pedrobaeza

### DIFF
--- a/mail_tracking/models/mail_message.py
+++ b/mail_tracking/models/mail_message.py
@@ -5,6 +5,7 @@
 from email.utils import getaddresses
 
 from odoo import api, fields, models
+from odoo.exceptions import AccessError
 from odoo.osv import expression
 from odoo.tools import email_split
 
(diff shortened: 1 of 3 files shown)


pedrobaeza (approved):
Code review

PR #266, opened 2026-09-10
[19.0][FIX] mail_tracking: do not swallow exceptions in send_email

### LINKED ISSUE #253: [19.0][18.0] mail_tracking: send_email() swallows every exception when a tracking record exists — emails are reported as sent when they were not
## Summary

`IrMailServer.send_email()` catches every exception and, **when a `mail.tracking.email` record exists for the message**, records the error on the tracking record and returns instead of re-raising — `mail_tracking/models/ir_mail_server.py` (lines 161-165 on `19.0`, 159-163 on `18.0`):

```python
        except Exception as e:
            if tracking_email:
                tracking_email.smtp_error(self, smtp_server_used, e)
            else:
                raise
```

Odoo core tells a successful send from a failed one **by the exception**, not by the return value, so returning normally tells core "this was delivered". Since this module creates a tracking record for *every* outgoing email (`mail_tracking/models/mail_mail.py::_prepare_outgoing_list`) and injects the pixel that `_build_email__` reads back as `X-Odoo-MailTracking-ID`, `tracking_email` is truthy for every normal HTML email: the `else: raise` branch is effectively dead, and core's failure handling is disabled database-wide.

**The visible consequence: emails that were never delivered are reported to users as delivered.**

## Measured behaviour

Odoo 19.0 CE (`19.0-20260723`, official image), fresh database, the SMTP session replaced by a stub so that valid recipients are really delivered and can be counted. Two partners are notified through the chatter (`message_post`, `mail.mt_comment`): one with an email address, one **without** — the case for which core raises `AssertionError(NO_VALID_RECIPIENT)`.

Same scenario, run on a database with `mail_tracking` installed and on a control database with core only:

| `mail.notification` of the recipient **without** email | with `mail_tracking` | control (core only) |
|---|---|---|
| `notification_status` | **`sent`** | `exception` |
| `failure_type` | **not set** | `mail_email_invalid` |

The valid recipient is delivered in both cases (a single SMTP envelope, `['valid@example.com']`) — so the module does not stop delivery here, it misreports it. Both orderings of the recipients give the same result.

Second scenario, `mail.mail` with `auto_delete=False` and the broken recipient **last** in `email_list` (verified by printing the list, not assumed):

| `mail.mail` after sending | with `mail_tracking` | control (core only) |
|---|---|---|
| `state` | **`exception`** | `sent` |
| `failure_type` | `unknown` | — |
| `failure_reason` | *"Error without exception. Probably due to sending an email without computed recipients."* | — |
| valid recipient delivered? | yes | yes |

So the same swallowed exception produces two opposite errors depending on position: a failed recipient in the middle is reported as delivered, and a failed recipient at the end marks the whole mail as failed even though everybody else received it.

## Why (core's loop)

`odoo/addons/mail/models/mail_mail.py::_send()`:

```python
res = None
...
for email in email_list:
    try:
        res = SendIrMailServer.send_email(msg, ...)
        if processing_pid:
            success_pids.append(processing_pid)
    except AssertionError as error:
        if str(error) == IrMailServer.NO_VALID_RECIPIENT:
            ...
            failure_type = "mail_email_invalid"
if res:  # mail has been sent at least once, no major exception occurred
```

1. `except AssertionError` never runs → a recipient without a valid address is never classified as `mail_email_invalid`.
2. The recipient is appended to `success_pids` as soon as `send_email` returns, whatever it returns; `_postprocess_sent_message()` then writes `notification_status = 'sent'` for a message that was never delivered.
3. `res` is overwritten on every iteration and the verdict is `if res:`, so a failure on the **last** recipient marks the whole mail as an exception.

The same asymmetry applies to any hard SMTP failure, not only to missing addresses.

Note that the module already anticipates this exact exception: `MailTrackingEmail.smtp_error()` has a dedicated branch for `str(exception) == IrMailServer.NO_VALID_RECIPIENT` setting `error_type = "no_recipient"`. The tracking record is annotated correctly — only core is left in the dark.

## Proposed fix

Always re-raise. The error is still recorded on the tracking record by the line above, and control returns to core, which can then classify the failure, keep delivering to the remaining recipients when it is a `NO_VALID_RECIPIENT`, and mark the mail for what it actually is.

```diff
         except Exception as e:
             if tracking_email:
                 tracking_email.smtp_error(self, smtp_server_used, e)
-            else:
-                raise
+            # Always re-raise: MailMail._send() tells success from failure by
+            # the exception, not by the return value. Returning here makes core
+            # mark the notification as 'sent' for a message that was never
+            # delivered, prevents NO_VALID_RECIPIENT from ever being classified
+            # as 'mail_email_invalid', and leaves `res` False so a failure on
+            # the last recipient marks the whole mail as an exception.
+            # The error is already recorded on the tracking record above.
+            raise
```

## Please read before merging

This is a behaviour change with a visible consequence: **failures that used to be silent will start surfacing as exceptions.** On the production database where we hit this, applying the fix immediately turned a backlog of previously hidden delivery failures into visible `exception` states — the tracking records had been sitting in `error` state all along, they were simply never propagated to the mails and notifications. That is the correct outcome, but it will look like a regression to anyone upgrading without warning, so it deserves a line in the migration notes.

Versions: `19.0.1.0.5` and `18.0.1.0.11` carry identical code. We are running this patch in production and are happy to open it as a PR if the approach is agreed.

skanndar:
Fixes #253

### The problem

`MailMail._send()` tells a successful delivery from a failed one by the **exception** raised by
`send_email()`, not by its return value. `mail_tracking` catches every exception and, when a
`mail.tracking.email` record exists, records the error on the tracking record and returns
normally, so core believes the mail was delivered:

* the recipient is appended to `success_pids` and `_postprocess_sent_message()` writes
  `notification_status = 'sent'` on a notification for a mail that never left — **an email that
  did not arrive is displayed to users as delivered, with no trace anywhere**;
* core's `except AssertionError` never runs, so a `NO_VALID_RECIPIENT` is never classified as
  `mail_email_invalid`;
* `res` stays falsy, so the mail keeps the placeholder failure reason core sets *before* the
  sending loop (*"Error without exception. Probably due to sending an email without computed
  recipients."*) with `failure_type = 'unknown'`, even when the other recipients were delivered.

A tracking record is created for *every* outgoing email
(`mail_tracking/models/mail_mail.py::_prepare_outgoing_list`), so this is not an edge case: with
`mail_tracking` installed, core's whole failure-handling path is disabled for the database.

### Not a duplicate of #215 / #212

#215 added `else: raise`, i.e. it re-raises only when there is **no** tracking record — the
branch that is effectively dead in practice. The case reported in #253 is the other one and is
still live on `19.0` (checked at `ee6a0e79`) and on `18.0`.

### The change

Re-raise unconditionally. The error is still recorded on the tracking record by the
`smtp_error()` call just above, so nothing is lost; only core's contract is restored.

### Tests

Two regression tests build the `mail.message` + `mail.mail` + `mail.notification` trio core uses
to decide whether a recipient was reached:

* `test_smtp_error_is_not_reported_as_sent` — fails without the patch with
  `AssertionError: 'sent' == 'sent'` (the notification of a mail that never left is marked as
  delivered) and, further down, `mail.failure_reason` is the placeholder instead of the real one;
* `test_smtp_error_still_tracks_the_error` — checks the tracking record is still annotated and
  that core can now classify the failure (`failure_type == 'mail_email_invalid'`), which fails
  without the patch with `AssertionError: 'mail_email_invalid' != 'unknown'`.

Module suite on `odoo:19` (19.0-20260817), fresh database: 46 tests / 0 failures before,
48 tests / 0 failures after.

### ⚠️ Note for adopters (worth a line in the changelog)

This is a behaviour change with a visible consequence: **delivery failures that used to be silent
now surface as `exception` states** on `mail.mail` and `mail.notification`. On a production
database with this patch, a backlog of previously hidden failures became visible at once. That is
the correct outcome — the tracking records had been in `error` state all along, the state was
simply never propagated — but it will look like a regression to anyone upgrading without warning.


pedrobaeza (commented):
The tests generated by AI are synthetic and are not bringing real flows, so better to remove them.

skanndar:
@pedrobaeza reworked the two tests to drive the real notification pipeline instead of building the message/mail/notification by hand: a message is posted to a partner and sent through `_notify_thread` with a failing SMTP, then the recipient's notification is asserted not to be `sent` (and a `NO_VALID_RECIPIENT` to be classified as `mail_email_invalid`). They still fail without the fix and pass with it. CI is green.

